PR #13, opened 2015-11-06
Fix Map and Set `forEach` method

nicksrandall:
In order to be consistent with the Map and Set spec, I have refactored
the forEach method on both collection primitives.

**THIS DOES INTRODUCE A BREAKING CHANGE**


nicksrandall:
See #13

mbostock:
Fixed in 0.2.0.

Issue #12, opened 2015-11-05
Map.forEach: arguments passed in wrong order.


nicksrandall:
The callback function for `Map.forEach` gets called with 3 arguments (in order) for each item in collection: 

1. Element Value
2. Element Key
3. Map

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/forEach

The `d3.map` gets called with key then value (so backwards).

I understand that you can't rely on browser support for native JavaScript collections (Map and Set) but I think you should mimic their apis as much as possible.

I'm happy to send in a PR with the change if you think such a change should be made.

nicksrandall:
FWIW, I think the spec is wrong and it should be your way. That said, lets follow standards.

mbostock:
Consistency with the standard is good, but there are complicating factors here, namely that d3.map is not a shim for ES6 Map because it’s not really possible to shim ES6 Map. You can get a lot closer than d3.map, but I taking it as far as you can would be counterproductive. Some differences:

* D3’s map coerces keys to strings, whereas ES6 Maps can use any object.
* D3 calls it [map.remove](https://github.com/d3/d3-arrays#map_remove), whereas ES6 says [map.delete](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/delete).
* D3 lacks a [map.clear](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/clear) method.
* D3’s map is not [iterable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/%40%40iterator). (There’s not really a way to shim this feature.)
* Likewise, D3’s [map.keys](https://github.com/d3/d3-arrays#map_keys) returns an array, not an iterator.
* Likewise, D3’s [map.values](https://github.com/d3/d3-arrays#map_values) returns an array, not an iterator.
* D3’s [map.entries](https://github.com/d3/d3-arrays#map_entries) method returns an array of {key, value} objects, not an iterator of [key, value].
* D3’s [map.set](https://github.com/d3/d3-arrays#map_set) returns the set value, rather than the map instance
* D3’s [map.size](https://github.com/d3/d3-arrays#map_size) is method, not a [read-only accessor property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/size).
[...]

PR #14, opened 2015-11-06
Rename {map,set}.forEach to {map,set}.each.

mbostock:
D3’s map.forEach method didn’t behave like ES6’s map.forEach anyway, because the order of arguments was reversed and it didn’t take a thisArg. So we might as well use our preferred name and make the API difference more apparent.

Related #12 #13.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,7 +283,7 @@ Returns an array of values for every entry in this map. The order of the returne
 
 Returns an array of key-value objects for each entry in this map. The order of the returned entries is arbitrary. Each entry’s key is a string, but the value has arbitrary type.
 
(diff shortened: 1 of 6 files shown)


nicksrandall:
I'm all for the name change. I agree that it will be beneficial to make the difference between d3.map and ES6 Map more apparent. However, I still think that the each method should pass params in order of value then key. Every other method in D3 seems to behave that way (data, index) as well as many JS similar JS methods. I'm happy to make the change if you agree with me.

mbostock:
Huh. I never thought about (data, index) being parallel to (value, key), but that does make sense. Changing the name at the same time as reversing the order should serve as a good forcing function for callers to fix their argument order.

nicksrandall:
You beat me to it. Thanks @mbostock!

mbostock:
Take a look at #16, too? Thanks!

PR #16, opened 2015-11-06
Collection improvements.

mbostock:
- [x] Add map.clear.
- [x] Add set.clear.
- [x] map.set should return the map, not the added value.
- [x] set.add should return the set, not the added value.


nicksrandall:
Looks good to me!

mbostock:
Thanks. Now wonder if map.remove, map.clear, set.remove and set.clear should also `return this` to enable chaining…

PR #21, opened 2015-12-02
Skewness and Kurtosis Calcuations

jeffcoukell:
This fork implements the features requested in issues https://github.com/d3/d3-arrays/issues/2 and https://github.com/d3/d3-arrays/issues/3.  Specifically, it implements the "sample skewness" formula from http://www.macroption.com/skewness-formula/, and the "sample excess kurtosis" formula from http://www.macroption.com/kurtosis-formula/.  These are the same formulas implemented by popular spreadsheet applications like Microsoft Excel and OpenOffice Calc.

Unit tests following the style of variance-test and deviation-test have been added for both skewness and kurtosis along with an implementation of inDelta from d3-random.


mbostock:
I appreciate the offer to implement these, but I think it’s best to wait until there’s demand for these functions. It may be more prudent to develop a separate module for more “advanced” statistics.

Issue #19, opened 2015-11-21
Histogram layout.


mbostock:
Seems like this repo could be a good place for it.

PR #23, opened 2015-12-03
Histogram!

mbostock:
I’m not totally sure about the histogram.thresholds API. It’s more natural when you’re specifying the thresholds, but it’s less natural when you’re specifying the number of bins (the number of thresholds plus one). But I think it’s probably fine.


mbostock:
Fixes #19.

PR #25, opened 2015-12-10
Allow threshold functions to return a count of thresholds.

mbostock:
This makes it easier to pass the count to [*linear*.ticks](https://github.com/d3/d3-scale#linear_ticks).

Edit: actually, we’ll just generate nice ticks automatically. Even better!


mbostock:
Related #26: we should choose nice, rounded thresholds when specified as a count.

PR #27, opened 2015-12-30
Add flatten.

mbostock:
Flatten is a recursive merge operator. For example, if you want to compute the extent of a stack layout, you can’t simply use extent because you need separate value accessors for the min and max. So, the obvious solution is to call min and max separately:

```js
[
  d3_array.min(series, function(s) { return d3_array.min(s, function(d) { return d[0]; }); }),
  d3_array.max(series, function(s) { return d3_array.max(s, function(d) { return d[1]; }); })
]
```

However, you could instead flatten the series into a single array and then compute the extent directly:

```js
d3_array.extent(d3_array.merge(d3_array.merge(series)))
```

Although more concise, this approach isn’t very efficient because you first create an array of N×M elements (N = number of series, M = number of points per series) and then copy it into an N×M×2 array.

The flatten method proposed in this pull request probably isn’t much more efficient. (It might even be slower because the length of the flattened array isn’t known beforehand, so the array can’t be pre-allocated. Instead, a queue of elements to visit is maintained during traversal. Also, this method will hang if the input array contains itself.)

But it is straightforward and concise, so it might be useful?

```js
d3_array.extent(d3_array.flatten(series))
```

Related note: I’m looking forward to being able to use generators and iterators.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,14 @@ Merges the specified *arrays* into a single array. This method is similar to the
 d3_array.merge([[1], [2, 3]]); // returns [1, 2, 3]
 ```
 
+<a name="flatten" href="#flatten">#</a> d3_array.<b>flatten</b>(<i>arrays</i>)
+
+Flattens the specified nested array of *arrays* into a single array. This method is essentially a recursive [merge](#merge) for each input element that is an array.
+
+```js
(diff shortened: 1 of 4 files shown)


mbostock:
https://jsperf.com/flatten-vs-merge-vs-minmax

This might be a bad idea. Or, I might be doing some premature optimization.

PR #35, opened 2016-06-08
Implement Kahan summation.

### LINKED ISSUE #33: Use Kahan summation algorithm?
h/t https://github.com/simple-statistics/simple-statistics/blob/master/src/sum.js

mbostock:
Fixes #33. Not sure if this is worth doing, though. I’d like some examples.

### DIFF
--- a/src/sum.js
+++ b/src/sum.js
@@ -1,16 +1,26 @@
 export default function(array, f) {
-  var s = 0,
+  var s0 = 0,
+      s1,
+      e = 0,
+      v,
       n = array.length,
-      a,
       i = -1;
 
   if (f == null) {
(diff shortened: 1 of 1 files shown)


Fil:
An example here https://observablehq.com/@fil/kahan-summation

mbostock:
Some precedent in numpy and Julia suggests that pairwise summation may be preferable (though unclear if using iterables will be performant):

https://en.m.wikipedia.org/wiki/Pairwise_summation

Fil:
I don't think pairwise summation solves the `1.0 + eps - eps` case.

Fil:
rebased in #149

PR #41, opened 2016-09-13
Extract accessor

geekplux:
drop some duplicate code in statistics methods.

### DIFF
--- a/src/extent.js
+++ b/src/extent.js
@@ -1,3 +1,5 @@
+import accessor from './accessor';
+
 export default function(array, f) {
   var i = -1,
       n = array.length,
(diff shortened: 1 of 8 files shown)


src/extent.js
```diff
@@ -1,24 +1,16 @@
...
-    }
-  }
-
-  else {
-    while (++i < n) if ((b = f(array[i], i, array)) != null && b >= b) { a = c = b; break; }
-    while (++i < n) if ((b = f(array[i], i, array)) != null) {
-      if (a > b) a = b;
-      if (c < b) c = b;
-    }
+  while (++i < n) if ((b = accessor(f, i, array)) != null && b >= b) { a = c = b; break; }
```
curran: This seems like it would slow performance, as it adds the overhead of a function invocation and a `f == null` comparison in a tight loop. It would be interesting to do some performance benchmarks of this code compared to the original.

curran: This seems like it would slow performance, as it adds the overhead of a function invocation and a `f == null` comparison in a tight loop. It would be interesting to do some performance benchmarks of this code compared to the original.

geekplux: thanks your advice. In the process of coding, I tried to use the `console.time()` and `console.timeEnd()` to watch the performance, and result was little difference between after and before. So..I guess whether the data amount in the test is so little?

geekplux:
and why not to use CI in d3's repo.. @mbostock

mbostock:
I appreciate the contribution, but I don’t see this as simpler, and without any empirical evidence to suggest one way is better than the other I’d rather keep things as-is. But thank you.

geekplux:
aha, no problem

PR #43, opened 2016-10-11
I think f may be a array.

CaoYouXin:


mbostock:
The `f` here is an *accessor* function; see the API reference for [d3.bisector](https://github.com/d3/d3-array/blob/master/README.md#bisector).

PR #44, opened 2016-10-18
Automatic fixes for d3/d3#2978

MatrixFrog:
This makes the Closure Compiler happy by not attempting to use variables before they're declared.

### DIFF
--- a/src/quantile.js
+++ b/src/quantile.js
@@ -2,13 +2,11 @@ import number from "./number";
 
 export default function(array, p, f) {
   if (f == null) f = number;
+  var n;
   if (!(n = array.length)) return;
   if ((p = +p) <= 0 || n < 2) return +f(array[0], 0, array);
   if (p >= 1) return +f(array[n - 1], n - 1, array);
-  var n,
-      h = (n - 1) * p,
-      i = Math.floor(h),
-      a = +f(array[i], i, array),
+  var h = (n - 1) * p, i = Math.floor(h), a = +f(array[i], i, array),
       b = +f(array[i + 1], i + 1, array);
   return a + (b - a) * (h - i);
 }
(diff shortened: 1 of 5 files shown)


MatrixFrog:
There will be fixes in 10 or so other repos. I was going to try to send all the pull requests automatically but there will probably formatting fixes that need to be done manually.

MatrixFrog:
You said on the referenced issue

> I am probably not willing to spend the necessary effort to rewrite the 50+ occurrences of this pattern to fix unsafe optimizations in the Closure Compiler

which I took to mean that you'd be happy to accept the changes if someone else sends the PRs. Is that understanding correct, and if so is there anything I can do to help get this PR accepted? Thanks.

mbostock:
I suppose I am not very motivated to incorporate this change because it adds to the arcanum I must remember when writing or maintaining code—I have to remember that I can’t just write standard JavaScript because there’s a downstream tool that doesn’t fully support the language. And so that wears me down a bit.

I know relying on *var* hoisting is questionable style, so I feel strange trying to justify it. But at the same time, this pattern is convenient if you want to group all your declarations into a single statement, and in JavaScript I feel like I’m always trying to write it as small as possible to save a few bytes in the client, even knowing that there’s a minifier to help out. 

I do appreciate your contribution and I do want to help, and I wouldn’t say I’m opposed to this change… it’s just I have a million other things I want to do, too, and many of them are more compelling to me personally, like the forthcoming release of TopoJSON 2.0 and d3-geo-projection 1.1. Part of me thinks, if you have an automated tool to fix the var hoisting, why not just run it on the code before you pass it to the Closure Compiler? Then you don’t need to patch the changes in the source. 😉

MatrixFrog:
Okay sounds good. Hopefully we can just fix the inliner in the compiler itself.

PR #51, opened 2017-02-08
Update README.md

dyst5422:
Just spent several hours looking into a bug that really came down to me not presorting the array passed to quantile. Wasn't until the 5th time I went back to read the documentation that I caught that it was a sorted array, so I thought I'd extend the emphasis on array to sorted array.


mbostock:
Sorry for the lost time. I’ll switch this to use bold (two asterisks) instead of italics, since in the D3 API Reference italics are typically used to indicate argument names.

PR #55, opened 2017-04-08
pairs checks that second arg is a function

1wheel:
Before c57a0848ff8bbf8c4cbc8fd544cb9c4ca4ca329f, this worked:

```
var lines = [
  [[10, 20], [20, 90], [30, 40]],
  [[10, 15], [20, 30], [30, 10]],
]

svg.appendMany(lines, 'g').appendMany(d3.pairs, 'path')
  .attrs({d: d => 'M' + d.join('L'), stroke: '#000'})
```

Now `d3.pairs` tries to use the `index` argument from `.data` as a function. Not a huge deal - `.appendMany(d => d3.pairs(d), 'path')` works - but it was nicer before.

### DIFF
--- a/src/pairs.js
+++ b/src/pairs.js
@@ -1,5 +1,5 @@
 export default function(array, f) {
-  if (f == null) f = pair;
+  if (typeof f !== "function") f = pair;
   var i = 0, n = array.length - 1, p = array[0], pairs = new Array(n < 0 ? 0 : n);
   while (i < n) pairs[i] = f(p, p = array[++i]);
   return pairs;


mbostock:
Sorry, but I try to use `== null` consistently to check for optional arguments; see d3/d3#2710 for some context. There’s a similar issue where I often want to do something like:

```js
["red", "green", "blue"].map(d3.rgb) // bad
```

But fortunately, as you mention, we have arrow functions now.

```js
["red", "green", "blue"].map(d => d3.rgb(d)) // good
```

PR #58, opened 2017-05-24
simplify max&min function

xuluxi:
Passed the test


mbostock:
The purpose of the current implementation is to avoid invoking an accessor function in the common case where that accessor function is the identity function. While it is simpler to supply the identity function as the accessor when one is not specified, it is slower (last time I checked), so it’s worth the additional complexity for a library.

Also, if we were to change this implementation, we’d need to change not just d3.min and d3.max but all the other functions in this library that similarly have optional accessors.

PR #59, opened 2017-07-12
Faster comparison for equality.

leodutra:
Faster and logic comparison for equality.


curran:
Have you any benchmarks that show the improvement in performance?

mbostock:
This fails for Date objects and other objects that implement valueOf for comparison.

PR #60, opened 2017-07-12
Faster equality comparison on ascending.js

leodutra:
Faster, and logic, equality comparison on ascending.js.


mbostock:
As with #59, this fails for Date objects and other objects that implement valueOf for comparison. In the future, please run the tests before submitting a pull request. Thank you!

PR #63, opened 2017-10-06
replace export default by normal export

luisvt:


mbostock:
Thanks, but the code seems fine as-is? Please propose reasons if you wish to discuss this change, as it would have ramifications across other D3 repositories and affect backwards-compatibility for anyone that is requiring these modules directly (rather than using the top-level index).

PR #68, opened 2018-04-17
Don’t filter explicit thresholds.

### LINKED ISSUE #54: Histogram merges last two values when last value coincides with the upper threshold
I wrote a [Stack Overflow question](http://stackoverflow.com/questions/43036419/why-is-d3-histogram-merging-the-last-two-bins-of-this-chart) about this, which documents the behaviour well. At the request of an answerer, I've created this issue to try and understand this behaviour better (or alternately register it as a bug if it is indeed such).

tl;dr — d3.histogram merges the last two bins when the last value coincides with the upper threshold. However, if you set `histogram.domain()` to `[extentMin -1, extentMax + 1]`, the problem seems to dissipate. 

Example code (courtesy of Gerardo Furtado):

```javascript
var data = d3.range(100);

const histogram = d3.histogram()
  .value(d => d)
  .thresholds(data);

var bins = histogram(data);

console.log(bins);
```

The last bin contains both 98 and 99, whereas the other bins only contain one value.

This isn't the case with the other thresholds:

```javascript
var data = d3.range(100);

const histogram = d3.histogram()
  .value(d => d)
  .thresholds(d3.thresholdFreedmanDiaconis(data, d3.min(data), d3.max(data)));

var bins = histogram(data);

console.log(bins)
```

```javascript
var data = d3.range(100);

const histogram = d3.histogram()
  .value(d => d)
  .thresholds(d3.thresholdScott(data, d3.min(data), d3.max(data)));

var bins = histogram(data);

console.log(bins)
```

```javascript
var data = d3.range(100);

const histogram = d3.histogram()
  .value(d => d)
  .thresholds(d3.thresholdSturges(data, d3.min(data), d3.max(data)));

var bins = histogram(data);

console.log(bins)
```

Any idea what's going on with this?

Thanks!

Kim-Ariel:
Only filter implicit thresholds by the domain. Fixes #54.

### DIFF
--- a/src/histogram.js
+++ b/src/histogram.js
@@ -14,6 +14,7 @@ export default function() {
   function histogram(data) {
     var i,
         n = data.length,
+        m,
         x,
         values = new Array(n);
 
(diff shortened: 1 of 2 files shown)


mbostock:
https://github.com/d3/d3-array/issues/54#issuecomment-294270741

Issue #72, opened 2018-08-23
Usage of top-level calls


Andarist:
Top level calls are hard to treeshake because tree-shakers/minifiers are often "afraid" of removing them because of potential side-effects of those calls.

Culprits:
https://github.com/d3/d3-array/blob/36edab36219b54812e833c136bc47dcab66b78fc/src/bisect.js#L4
https://github.com/d3/d3-array/blob/36edab36219b54812e833c136bc47dcab66b78fc/src/ticks.js#L1-L3

Possible solution:
Add `#__PURE__` annotations before those calls to reassure UglifyJS that those calls can be dropped if their results stay unused - either manually or with help of https://github.com/Andarist/babel-plugin-annotate-pure-calls

Intent to implement:
yes - would only have to know if you want to tackle this one

mbostock:
I’ll tackle this when Rollup adds support for pure annotations rollup/rollup#1293.

conartist6:
I just opened https://github.com/rollup/rollup/pull/2429 against Rollup adding that functionality. Feel free to weigh in.

PR #73, opened 2018-08-23
Add sideEffects: false to pkg.jsoj

Andarist:
Partially helps with https://github.com/d3/d3-array/issues/72 but only for webpack@4+


mbostock:
Closing to inactivity.

PR #78, opened 2018-11-10
Update rollup.js ->  @1.3.0 cause webpack UglifyJs  build error

jeoy:
There is es6 syntax `let` and `const` in this file,  cause Webpack build error when d3-array@1.3.0 used in project.

> [ Error: static/js/vendor.169496a4301f622e0093.js from UglifyJs


mbostock:
Duplicate of #80.

PR #84, opened 2018-11-11
Fix cross for iterables which can be only iterated once

jviide:
The documentation states that cross returns the Cartesian product of the specified iterables. Some iterables, such as generators, can be iterated through only once. The current implementation of cross returns invalid results:

```js
const { cross } = require("d3-array");

function *gen(...values) {
  yield *values;
}

cross(gen(1, 2), gen("x", "y"));  // [ [ 1, 'x' ], [ 1, 'y' ] ]
```

This pull request adds a new test and fixes cross by first turning non-array iterable arguments into arrays. As a small optimization array arguments are used as-is.

### DIFF
--- a/src/cross.js
+++ b/src/cross.js
@@ -1,6 +1,9 @@
 function* product(head, ...rest) {
   if (rest.length) {
-    const tail = rest.pop();
+    let tail = rest.pop();
+    if (!Array.isArray(tail)) {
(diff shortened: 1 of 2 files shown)


mbostock:
Fixed in 2.0.1. Thank you!

https://github.com/d3/d3-array/compare/v2.0.0...v2.0.1

PR #85, opened 2018-11-11
Fix cross for zero arguments and zero-length arguments

jviide:
The current master version of cross loops infinitely when there are no arguments or any of the arguments is zero-length. This pull request adds a couple of checks to handle such cases as well as three tests for future reference.


mbostock:
Heh, thanks for the tests! I’ll get back to this after lunch.

PR #94, opened 2019-03-07
accept iterable indexes and return same type of array

### LINKED ISSUE #92: d3.permute should inherit the type of the passed-in array.
For example, if d3.permute a Int32Array, you should get back a new Int32Array.

### LINKED ISSUE #93: d3.permute should accept an iterable for the order.
Currently it requires *indexes* to be an array, but this is overly strict.

Fil:
fixes #92
fixes #93

### DIFF
--- a/src/permute.js
+++ b/src/permute.js
@@ -1,5 +1,13 @@
 export default function(array, indexes) {
   var i = 0, permutes = array.slice();
-  for (var p of indexes) permutes[i++] = array[p];
+  for (var p of indexes) {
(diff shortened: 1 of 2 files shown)


Fil:
Oops. Doesn't work for the (untested) case where the index is longer than the typed array 😫. New version coming.

mbostock:
I closed #92. For #93 I think we want something like this:

```js
function permute(source, keys) {
  return Array.from(keys, key => source[key]);
}
```

PR #104, opened 2019-04-30
fix #97 quantileSorted

### LINKED ISSUE #97: d3.quantile is difficult to use
Although it's properly documented, it's easy to fall in the trap of forgetting to sort & filter the data, and a bit difficult to find the correct way to do it.

Shouldn't we have a slower but foolproof d3.quantile, and a fast d3.quantileRaw (or any other name) for the few people who know what they are doing?

Another way to look at this is that the reason for this behavior is to optimize when extracting several quantiles ; but then this could be rewritten as d3.quantile(values, fractions, accessor) where fractions is a list.

Fil:
quantile accepts unsorted iterables; quantileSorted expects a sorted array

closes https://github.com/d3/d3-array/issues/97

### DIFF
--- a/src/quantile.js
+++ b/src/quantile.js
@@ -1,16 +1,10 @@
-import number from "./number.js";
-import identity from "./identity.js";
+import {default as number, numbers} from "./number.js";
 
 export default function quantile(values, p, valueof = number) {
-  values = Array.from(values)
-    .map((d, i) => valueof(d, i, values))
+  values = Float64Array.from(numbers(values, valueof))
     .filter(v => !isNaN(v))
     .sort((a, b) => a - b);
-  var n = values.length;
-  if (!n) return;
-  if ((p = +p) <= 0 || n < 2) return +values[0];
-  if (p >= 1) return +values.pop();
-  return quantileSorted(values, p, identity);
+  return quantileSorted(values, p);
 }
 
 export function quantileSorted(values, p, valueof = number) {
--- a/src/quantile.js
+++ b/src/quantile.js
@@ -1,6 +1,19 @@
 import number from "./number.js";
+import identity from "./identity.js";
 
 export default function quantile(values, p, valueof = number) {
+  values = Array.from(values)
+    .map((d, i) => valueof(d, i, values))
+    .filter(v => !isNaN(v))
+    .sort((a, b) => a - b);
+  var n = values.length;
+  if (!n) return;
+  if ((p = +p) <= 0 || n < 2) return +values[0];
+  if (p >= 1) return +values.pop();
+  return quantileSorted(values, p, identity);
+}
+
+export function quantileSorted(values, p, valueof = number) {
   if (!(n = values.length)) return;
   if ((p = +p) <= 0 || n < 2) return +valueof(values[0], 0, values);
   if (p >= 1) return +valueof(values[n - 1], n - 1, values);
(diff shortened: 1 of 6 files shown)


src/quantile.js
```diff
@@ -1,6 +1,19 @@
 import number from "./number.js";
+import identity from "./identity.js";
 
 export default function quantile(values, p, valueof = number) {
+  values = Array.from(values)
+    .map((d, i) => valueof(d, i, values))
+    .filter(v => !isNaN(v))
```
mbostock: This should use the `Float64Array.from(numbers(values, valueof))` technique from median.

src/quantile.js
```diff
@@ -1,6 +1,19 @@
...
+    .sort((a, b) => a - b);
+  var n = values.length;
+  if (!n) return;
+  if ((p = +p) <= 0 || n < 2) return +values[0];
+  if (p >= 1) return +values.pop();
```
mbostock: These lines can be dropped?

```js
var n = values.length;
if (!n) return;
if ((p = +p) <= 0 || n < 2) return +values[0];
if (p >= 1) return +values.pop();
```

They appear to do the same thing as quantileSorted would do.

src/quantile.js
```diff
@@ -1,6 +1,19 @@
...
+  return quantileSorted(values, p, identity);
```
mbostock: I think I would simplify and not pass `identity` here, since the default `number` will do just as well, and will only be called at most twice.

src/quantile.js
Fil: yes. moved numbers* to number.js to mutualize it

src/quantile.js
Fil: true

Fil:
> Does that sort of performance change still qualify as backwards-compatible, in your view?

I understand how it could be seen as a regression. In my tests the cost is 1s for 1 million numbers. (The fact that the array is already sorted doesn't seem to lower that cost much.)

src/quantile.js
mbostock: You don’t need to filter since numbers will only give you not-NaN numbers.

mbostock:
Superseded by #120.

mbostock:
I’ve merged and released this, but would it be faster to always use quickselect rather than sort?

Fil:
From my limited tests on a 1e6 array, sort looks 8 times slower than quickselect.

So I suppose in that case the answer is in favor of quickselect only when querying fewer than 8 quantiles.

But if you work on 1e7 numbers, sort takes forever and quickselect returns much faster. So for *large* arrays this would indicate a definite bonus on quickselect.

https://observablehq.com/d/95cfea984ca7d445

mbostock:
You could always use quantileSorted if you want to compute multiple quantiles. I’m mainly interested in optimizing the case of computing a single (or a few) quantiles on unsorted data. An 8x improvement sounds highly worthwhile.

Fil:
Yes for d3.quantile it seems like it's always a win. Up to the point that we shouldn't recommend using quantileSorted unless you need to query it a high number of times (>8) — or the values are already sorted for some other reason.

I didn't factor in the max (and min if n is odd) you need on top, but that's probably a negligible cost.

An updated quantile in PR #121.

Now I'm having second thoughts re: quantileSorted: does it need to have an accessor?

Issue #109, opened 2019-05-20
d3.least[Index] should promote single-argument accessors to comparators.


mbostock:
Like d3.bisector does.

https://github.com/d3/d3-array/pull/108#issuecomment-493894485

mbostock:
Fixed in #111.

PR #110, opened 2019-05-21
least+leastIndex with accessor 

Fil:
fixes https://github.com/d3/d3-array/issues/109

It's a lot of code. Using `v = valueof ? valueof(value) : value` would be maybe a weeeny bit slower, but take less lines?


Fil:
It would simplify some examples in  https://observablehq.com/@d3/d3-least and d3.path

mbostock:
Superseded by #111. (You have write access to this repo, don’t you?)

Fil:
> You have write access to this repo, don’t you?

nope

mbostock:
Okay, I’ve created a team with write access to all D3 repos and added you!

PR #121, opened 2019-08-22
use quicksort rather than sort in quantile

Fil:
follows the discussion at https://github.com/d3/d3-array/pull/104#issuecomment-523732923

### DIFF
--- a/src/quantile.js
+++ b/src/quantile.js
@@ -3,7 +3,7 @@ import min from "./min.js";
 import quickselect from "./quickselect.js";
 import number, {numbers} from "./number.js";
 
-export default function quantile(values, p, valueof = number) {
+export default function quantile(values, p, valueof) {
   values = Float64Array.from(numbers(values, valueof));
   if (!(n = values.length)) return;
   if ((p = +p) <= 0 || n < 2) return min(values);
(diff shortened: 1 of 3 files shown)


src/quantile.js
```diff
@@ -1,8 +1,23 @@
...
+  if (p >= 1) return max(v);
+  var n,
+      i = (n - 1) * p,
+      i0 = Math.floor(i);
+  quickselect(v, i0);
+  var value0 = -Infinity, value1 = +Infinity;
+  for (var j = 0; j < n; j++) {
+    if (j <= i0 && v[j] > value0) value0 = v[j];
+    if (j > i0 && v[j] < value1) value1 = v[j];
+  }
```
mbostock: What about `min(v.subarray(0, i0))` and `max(v.subarray(i0, n))` or similar?

Fil: :blush:

src/quantile.js
```diff
@@ -1,10 +1,19 @@
-import ascending from "./ascending.js";
+import max from "./max.js";
+import min from "./min.js";
+import quickselect from "./quickselect.js";
 import number, {numbers} from "./number.js";
 
-export default function quantile(values, p, valueof) {
+export default function quantile(values, p, valueof = number) {
```
mbostock: This default value `= number` isn’t needed because `numbers(values, undefined)` will be equivalent, no? I’ll try it and see.

Issue #124, opened 2019-10-16
Median function inefficiently calls quantile method (instead of quantileSorted)?


jheer:
The [median implementation](https://github.com/d3/d3-array/blob/master/src/median.js#L12) performs an array copy, runs quickselect, and then invokes `quantile`... which in turn performs an array copy, sorts the array, and then calls `quantileSorted`. This appears to be doing costly unnecessary extra work.

Should the `median` method be calling `quantileSorted` instead?

mbostock:
Derp! Thanks for finding and fixing this.

mbostock:
Fixed in #121.

PR #125, opened 2019-10-16
Update median to use quantileSorted.

jheer:
Fixes d3/d3-array#124.


Fil:
My bad— in #120 I forgot to check how quantile was used. However I'd rather fix this with #121 which uses quicksort in quantile, and (now) uses quantile(.5) for the median.

mbostock:
I agree with @Fil. Superseded by #121 which is now merged. Thank you both!

Issue #102, opened 2019-04-21
d3.cumsum?


yurivish:
Taking a cumulative sum is a common function in numeric computing and I've found myself repeatedly reaching for it over the years. 

```
[1, 2, 3, 4].reduce((a, b) => (a.push(b + (a.length ? a[a.length-1] : 0)), a), [])
```

It can be implemented in one line, though you can achieve better performance by copying the array and running through `1:n` adding `a[n-1]` to `a[n]` inside each iteration.

mbostock:
I think I would implement this as

```js
function cumsum(values) {
  let sum = 0;
  return Float64Array.from(values, v => sum += v);
}
```

so that it works with iterables. But I’d also change it to coerce to numbers, treat NaN as zero, and take an accessor.

Fil:
like so?
~~~{javascript}
function cumsum(values, valueof) {
  var sum = 0, index = 0, value;
  return valueof === undefined
    ? Float64Array.from(values, v => (sum += +v || 0))
    : Float64Array.from(values, v => (sum += +valueof(v, ++index, values) || 0));
}
~~~

https://observablehq.com/@fil/cumsum

Fil:
Documentation: https://observablehq.com/d/918b60f6963a81e1

mbostock:
Added in 2.4.

PR #126, opened 2019-10-18
cumsum

Fil:
solves https://github.com/d3/d3-array/issues/102


Fil:
I merged it but need you to push it as v2.4.0 to npm.

PR #149, opened 2020-06-11
Implement Kahan summation.

Fil:
rebased, supersedes #35 
fixes https://github.com/d3/d3-array/issues/33

speed is not bad (maybe ~10% slower according to https://observablehq.com/d/723090bd2ef54f5f)

### DIFF
--- a/src/sum.js
+++ b/src/sum.js
@@ -1,18 +1,22 @@
 export default function sum(values, valueof) {
-  let sum = 0;
+  let sum0 = 0, sum1, error = 0;
   if (valueof === undefined) {
     for (let value of values) {
       if (value = +value) {
-        sum += value;
+        sum1 = sum0 + (value -= error);
+        error = sum1 - sum0 - value;
+        sum0 = sum1;
       }
     }
   } else {
     let index = -1;
     for (let value of values) {
       if (value = +valueof(value, ++index, values)) {
-        sum += value;
+        sum1 = sum0 + (value -= error);
+        error = sum1 - sum0 - value;
+        sum0 = sum1;
       }
     }
   }
-  return sum;
+  return sum0;
 }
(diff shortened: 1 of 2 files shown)


Fil:
(if the speed hit is not acceptable we can expose it as kahansum(), or just leave it as an example on observable)

mbostock:
From https://numpy.org/doc/stable/reference/generated/numpy.sum.html:

> For floating point numbers the numerical precision of sum (and np.add.reduce) is in general limited by directly adding each number individually to the result causing rounding errors in every step. However, often numpy will use a numerically better approach (partial pairwise summation) leading to improved precision in many use-cases. This improved precision is always provided when no axis is given. When axis is given, it will depend on which axis is summed. Technically, to provide the best speed possible, the improved precision is only used when the summation is along the fast axis in memory. Note that the exact precision may vary depending on other parameters. In contrast to NumPy, Python’s math.fsum function uses a slower but more precise approach to summation. Especially when summing a large number of lower precision floating point numbers, such as float32, numerical errors can become significant. In such cases it can be advisable to use dtype=”float64” to use a higher precision for the output.

From https://docs.python.org/3/library/math.html#math.fsum:

> Return an accurate floating point sum of values in the iterable. Avoids loss of precision by tracking multiple intermediate partial sums… The algorithm’s accuracy depends on IEEE-754 arithmetic guarantees and the typical case where the rounding mode is half-even. On some non-Windows builds, the underlying C library uses extended precision addition and may occasionally double-round an intermediate sum causing it to be off in its least significant bit.

Lastly, https://code.activestate.com/recipes/393090/

My inclination is that we should probably leave d3.sum as-is, and then mirror Python’s math.fsum implementation in a new d3.fsum?

Fil:
I've implemented fsum but I'm yet to find a series that would fault kahan summation (I always have the extact same result with fsum and kahan). And the speed it twice slower.
https://observablehq.com/d/723090bd2ef54f5f

Fil:
Closing in favor of fsum #152

Issue #90, opened 2019-01-28
rangeInclusive?


curran:
I frequently find myself using [d3.range](https://github.com/d3/d3-array#range) and desiring an inclusive range. I end up writing code like this:

```js
d3.range(min, max, step).concat(max)
```

Since this appears rather cryptic, I thought I'd propose this clearer alternative:

```js
d3.rangeInclusive(min, max, step)
```

This comes up at least once a month in my day to day work, I wonder if this is true for others as well.

Anyone else think this would be a useful addition to `d3-array`?

Fil:
I usually do something like `d3.range(min, max+0.001)` :)

Fil:
Closing this issue due to inactivity. Please do not hesitate to add a comment or reopen the issue if you think it is still worth pursuing.

PS: I think I like `d3.range(min, max, step).concat(max)` better, because it's more explicit (one less documentation item to know). Also I'd favor documenting d3.ticks(0,1,1000) which covers half the use cases.

mbostock:
We could also introduce a linspace method as in numpy. It functions more like d3.ticks, except it doesn’t attempt to produce human-readable values.

https://numpy.org/doc/stable/reference/generated/numpy.linspace.html

PR #150, opened 2020-06-11
Add d3.linspace.

mbostock:
Related #90.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -24,6 +24,7 @@ export {default as permute} from "./permute.js";
 export {default as quantile, quantileSorted} from "./quantile.js";
 export {default as quickselect} from "./quickselect.js";
 export {default as range} from "./range.js";
+export {default as linspace} from "./linspace.js";
 export {default as least} from "./least.js";
 export {default as leastIndex} from "./leastIndex.js";
 export {default as greatest} from "./greatest.js";
--- a/src/linspace.js
+++ b/src/linspace.js
@@ -1,0 +1,12 @@
+export default function(start, stop, n = 50, endpoint = true) {
+  n = Math.floor(n), start = +start, stop = (stop - start) / (endpoint ? n - 1 : n);
+
+  var i = -1,
+      range = new Array(n);
+
+  while (++i < n) {
+    range[i] = start + i * stop;
+  }
+
+  return range;
+}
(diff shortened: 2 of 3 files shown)


mbostock:
Maybe worth adding [logspace](https://numpy.org/devdocs/reference/generated/numpy.logspace.html) and [geomspace](https://numpy.org/devdocs/reference/generated/numpy.geomspace.html) too, while we’re here?

Fil:
I can see how it would be useful, esp. if it returned a Float64Array (or had some option like `dtype`), and satisfy #90 at the same time.

Not sure about floor(n), since it could be nice to continuously change *n* (in the case where * endpoint* is false).

For logspace and geomspace it looks like there is no particular algorithmic challenge and not much use. I wouldn't want to add weight (bytes and documentation and cognitive load) just for the sake of it :)

curran:
Very nice!

curran:
This looks like a nice PR. Could be merged I suppose.

mbostock:
I don’t think the positional arguments with defaults are appropriate here because they aren’t analogous to Python’s named arguments. For example in Python you would say stuff like

```python
np.linspace(2.0, 3.0, 5)
np.linspace(2.0, 3.0, num=5)
np.linspace(2.0, 3.0, num=5, endpoint=False)
np.linspace(2.0, 3.0, endpoint=False)
```

and the meaning is fairly clear. However in JavaScript it would be

```js
d3.linspace(2, 3, 5)
d3.linspace(2, 3, 5)
d3.linspace(2, 3, 5, false)
d3.linspace(2, 3, undefined, false)
```

which I think is much less obvious. So I wonder if instead we should do something like:

```js
d3.linspace(2, 3, 5)
d3.linspace(2, 3, {num: 5})
d3.linspace(2, 3, {num: 5, endpoint: false})
d3.linspace(2, 3, {endpoint: false})
```

This would also allow us to create typed arrays:

```js
d3.linspace(2, 3, {num: 5, type: Float64Array})
```

I also don’t really like feeling stuck with Python’s names for these options. I think I’d prefer `n` or `length` instead of `num`. I’m using `type` instead of `dtype` because in JavaScript it would be the type of the array rather than the type of the value. (And I thought about `inclusive` or `exclusive` instead of `endpoint` but I couldn’t make up my mind.)

And lastly I wonder if there should be more symmetry between d3.range and d3.linspace. Like, what about:

```js
d3.range(2, 3, {length: 5, inclusive: true})
```

Also this feels fairly low priority and ambiguous so I’m not very motivated to think about this too much. 🤷 Sorry!

Issue #56, opened 2017-04-12
Histogram Blur / Smoothing


curran:
It would be useful to have a "blur" function to approximate kernel density estimation.

This function would accept the output from `histogram`, conceptually looking something like:

```js
blur(histogram(data))
```

Original discussion in https://github.com/d3/d3-contour/issues/7#issuecomment-293321567

The implementation could be similar to [blurX or blurY in d3-contour](https://github.com/d3/d3-contour/blob/2fd7927e8c3ba2a7cf01bc0688ae0b2d2c1bf6ae/src/blur.js#L4).

This article may be relevant: [Convolve n Square Pulses to Gaussian](https://oscar6echo.blogspot.in/2012/10/convolve-n-square-pulses-to-gaussian.html).

curran:
Mind blown after seeing this

>Due to the central limit theorem, the Gaussian can be approximated by several runs of a very simple filter such as the moving average. The simple moving average corresponds to convolution with the constant B-spline ( a rectangular pulse ), and, for example, four iterations of a moving average yields a cubic B-spline as filter window which approximates the Gaussian quite well.

from [Wikipedia: Gaussian filter](https://en.wikipedia.org/wiki/Gaussian_filter).

Makes me wonder if we could add moving average functionality to d3-array as well, then use that to implement `blur`.

curran:
I did an experiment and found that moving averages do seem to converge on what looks like a Gaussian.

[![image](https://cloud.githubusercontent.com/assets/68416/24948503/1f181f28-1f88-11e7-89ca-9050a0792fad.png)](https://bl.ocks.org/curran/853fa00b8f0732fb2bee7fccfd7b4523)
https://bl.ocks.org/curran/853fa00b8f0732fb2bee7fccfd7b4523

Also related to https://github.com/d3/d3-shape/issues/43

curran:
I keep coming across scenarios where a bit of smoothing would be useful, for example this one:

![image](https://user-images.githubusercontent.com/68416/35920761-c64326fa-0c3e-11e8-98ae-eaf7acf4879c.png)

[CO2 Emissions Stacked Area Chart](https://datavis.tech/vis/3f5e0ba76cbf405e9fd87201bf21cac6)

curran:
Proposed implementation:

```js
const blur = (data, property) => data.map((d, i) => {
  const previous = (i === 0) ? i : i - 1;
  const next = (i === data.length - 1) ? i : i + 1;
[...]

PR #151, opened 2020-06-13
d3.blur

Fil:
fixes https://github.com/d3/d3-array/issues/56

see https://observablehq.com/@fil/moving-average-blur

- [x] remove iterations from the public API (keep it only for debugging)
- [x] <del>stackblur algo (a bit faster, but more difficult to understand)</del>
- [x] consistent ES6
- [x] verify that it can be used by d3-contour, in particular:
  - [x] allow different radius on x/y (https://github.com/d3/d3-contour/issues/27)
  - [x] the shape around a single dot should be circular(ish)
- [x] add fractional blurring (radius = .5)


curran:
Amazing!

Fil:
I'm pushing a new version which reduces the problem of the extremities being "dampened" by duplicating (*radius* times) the value at each extremity.

Bonus, this approach makes it even faster (we're now under 50% of the original time).

before:
before1
after:
after1



On the random image and a large radius, it creates a visible artifact: the values on the edges tends to impose themselves. Not strictly better or worse than the original, only different. (The solution when this is a problem is to compose first a small blur radius, then a larger one.)

before:
before
after:
after

curran:
Fascinating. Yeah I thought there was something going on with the extremities being "dampened", with the Gaussian blur approach in general. It's great that your implementation accounts for this!

Some interesting related reading that might touch upon this issue:

 * [Discrete Gaussian](https://en.wikipedia.org/wiki/Gaussian_function#Discrete_Gaussian)
 * [The discrete Gaussian kernel](https://en.wikipedia.org/wiki/Scale_space_implementation#The_discrete_Gaussian_kernel)

src/blur.js
```diff
@@ -0,0 +1,64 @@
+function blurTransfer(V, r, n, vertical) {
+  const [source, target] = V,
+    m = Math.floor(source.length / n),
+    w = (r << 1) + 1,
```
curran, line 4: Is this an implementation of a well known formula/algorithm? Anywhere I could read up on this math? I'm finding it a bit difficult to follow. Thanks!

README.md
```diff
@@ -536,6 +536,37 @@ Returns an array of arrays, where the *i*th array contains the *i*th element fro
...
 
+### Blur
+
+<a name="blur" href="#blur">#</a> d3.<b>blur</b>() · [Source](https://github.com/d3/d3-array/blob/master/src/blur.js), [Examples](https://observablehq.com/@d3/d3-blur)
+
+Creates a blur transformer, which can blur (or smooth) an *array* of values by repeated application of a moving average transform.
+
+<a name="blur_radius" href="#blur_radius">#</a> *blur*.<b>radius</b>([*radius*])
+
+If *radius* is specified, sets the radius of the transformation: on each iteration, the value of a point is transformed into the mean value of itself and the *radius* points of data surrounding it on the left and on the right (taking into account the edges). If *radius* is not specified, return the current radius. Defaults to 5.
```
curran, line 547: Any particular reason it defaults to 5?

src/blur.js
curran: Reading up on left shift now... https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#Left_shift

curran: Looks like this is equivalent to `w = r * 2 + 1`? Is the `<<` a performance optimization?

src/blur.js
```diff
@@ -0,0 +1,64 @@
...
+      const k = ki * i + kj * j;
+      sr += source[Math.min(k, kn)] - source[Math.max(k - W, k0)];
+      target[Math.max(k - R, k0)] = sr * w1;
+    }
+  }
+  V.reverse();
+}
+
+export default function blur() {
+  var radius = 5,
```
curran, line 23: Use of `var` vs. `let`/`const` is inconsistent within this file.

src/blur.js
Fil: It also performs coercing the value to integer. Note that this was in the original code; it doesn't bring much in terms of performance since it's only called once.

README.md
Fil: It's the current default in the d3-contour code, but it will change if we switch to the stackblur algorithm (in which radius = 7, iterations=1 gives the same "amount of blurring" as radius=5 with 3 iterations).

src/blur.js
Fil: Yeah it's going to be inconsistent all over the d3 code base for a while. But for "new" code I suppose we'll want ES6?

Fil:
Another use case for this type of blurring: "hue" blurring on an image. We first read each pixel's value in the Lab model (or HCL, HSL…), then blur a and b, and reassemble the image:
https://observablehq.com/d/b729443cd180e56f

Fil:
Finally I don't think the rgba thing is in the scope of this pull-request. It creates a bit of complications (stride, channel, dtype) "just" for specific optimizations, and it's possible to do without.

Fil:
I've added a notion that the shape created around a dot (or single pixel) should be circular(ish), so that d3-contours are more "round" than "square". Unfortunately it seems that the stackblur's distribution profile results in shapes that are a bit squarish:

simple sum * 3 iterations
sum blur

sum of sums
sumsum (stack) blur

So I think I have to either go back on that part, offer the two possibilities, or implement it separately in d3-contours.

Fil:
I went back to the original algorithm (to have "circular" shapes for d3-contour), and introduced non-integer radius, which help blur "just a little", for example with blur().radius(0.1).

I feel this PR is now complete.

curran:
Wow, this is awesome! Thank you @Fil for your work on this.

Fil:
Another demo with the "cloud contours" notebook: https://observablehq.com/d/c9d89106eafefb4f

curran:
Wow that is so cool!

src/blur.js
```diff
@@ -0,0 +1,88 @@
+import {floor, max, min} from "./math.js";
```
curran, line 1: Minor simplification opportunity: remove `src/math.js` and change this to:

```suggestion
const {floor, max, min} = Math;
```

README.md
```diff
@@ -536,6 +536,39 @@ Returns an array of arrays, where the *i*th array contains the *i*th element fro
 d3.zip([1, 2], [3, 4]); // returns [[1, 3], [2, 4]]
 ```
 
+### Blur
+
+<a name="blur" href="#blur">#</a> d3.<b>blur</b>() · [Source](https://github.com/d3/d3-array/blob/master/src/blur.js), [Examples](https://observablehq.com/@d3/d3-blur)
```
curran, line 541: I notice the examples link does not resolve https://observablehq.com/@d3/d3-blur

test/inDelta.js
```diff
@@ -0,0 +1,26 @@
+var tape = require("tape");
```
curran, line 1: Surprising to see `var`.

test/inDelta.js
```diff
@@ -0,0 +1,26 @@
...
+    expected: expected
+  });
+};
+
+function inDelta(actual, expected, delta) {
+  return (Array.isArray(expected) ? inDeltaArray : inDeltaNumber)(actual, expected, delta);
+}
+
+function inDeltaArray(actual, expected, delta) {
+  var n = expected.length, i = -1;
```
curran, line 18: Surprising to see `var`.

src/blur.js
Fil: Yep, I'm not sure why in D3 some math functions are abstracted like this, and others are not.

README.md
Fil: I'll publish some examples if the PR lands

test/inDelta.js
Fil: that file was just copied from other d3 repos :)

README.md
Fil: thanks for the review curran!

curran: My pleasure! Thanks for your great work on this feature.

Fil:
I find myself using this code more often than I thought. For example, when doing image FFT, to display the contours of a magnitude chart (which is too spiky if you take the raw values).

curran:
Nice. This PR is super cool. Can't wait for it to be merged!

Is there any more work needed here to complete the PR? Looks ready to me.

curran:
If this could improve `d3-contour`, _and_ provide new generic blur functionality independent of `d3-contour`, could be a big win for D3 on the whole.

mbostock:
This is a somewhat superficial reaction, but I wonder what this would look like without the fluent-style API. So, instead of:

```js
const blurred = d3.blur()
      .radius(radius)
      .width(detail.width)
      .value(d => d.b)
  (pixels);
```

You might say:

```js
const blurred = d3.blur(pixels, radius, detail.width, d => d.b);
```

The other thing that feels weird is that d3-array methods are, generally, 1-dimensional methods. Here d3.blur is two-dimensional, expressed in terms of *x* and *y* (or horizontal and vertical). This makes me think that this method should perhaps live in a different module that’s more related to 2D images. Alternatively, would it be possible to have a slightly lower-level API with an optional [*stride*](https://en.wikipedia.org/wiki/Stride_of_an_array) argument instead (where *stride* = 1 for “horizontal” and *stride* = *width*? for vertical), similar to [ndarray](https://github.com/scijs/ndarray)?

mbostock:
Another consideration is whether d3.blur should happen in-place, and eliminate the accessor. Would that be possible or does it require a copy?

```js
const blurred = Float64Array.from(pixels, d => d.b);
d3.blur(blurred, radius, detail.width);
```

Fil:
It's also 1-d, with a no-op vertical blur in that case. Internally there are two copies, that we swap at each iteration, so I'm not sure it can be done "in place", though maybe by tracing the parity of swaps it might be possible to save 1 copy allocation. In practice though I think I prefer this "immutable" version.

(As I needed it on several experiments I was just at this instant publishing it on github and npm (as array-blur) — but happy to revert that if it can be integrated into D3 in some way.)

curran:
Thrilled to see that this was published in https://github.com/Fil/array-blur !

cdaringe:
traversing thru the observablehq forums lead the thru the d3 docs and back to here. was looking for EMWA or movingAverage for 1d data, and found array-blur instead (which is actually more appropriate for my use case anyway :)). great work fil! are the outstanding blockers around whether or not such functionality is relevant/welcome here, or rather outstanding qualms with the interface? just curious :)

Fil:
Thank you, Christopher!

While it may never be included in d3-array, you can use this function by calling the [array-blur](https://github.com/Fil/array-blur) module. I've packaged it just like the d3 modules, which means you can even attach it to the d3 symbol if you like, with:
```js
const d3 = require("d3@7", "array-blur@1");
const blur = d3.blur().radius(1.3).width(w);
blur(pixels); // blurred image
```

Another possibility is that it would exist as a stand-alone module d3-blur (_i.e._ a renaming of array-blur); this would keep this 2D notion separate from d3-array, and allow d3-contour to use this faster and better blurring method.

I've made this notebook to summarize various use cases, in particular how to use it with Observable Plot (by passing _blur().radius(r)_ to Plot.mapY):
https://observablehq.com/@fil/array-blur-d3-plot

Fil:
We'll need two documentation notebooks:
* https://observablehq.com/d3/d3-blur (we can show the profile of the kernel in 1d and 2d, also how to use it on a grayscale data matrix)
* https://observablehq.com/d3/d3-blurimage (showing image blur and laplacian)

src/utils.js
```diff
@@ -0,0 +1,3 @@
+export function arrayify(values) {
```
mbostock, line 1: I think this file is no longer needed.

mbostock (changes requested):

mbostock (approved):

PR #154, opened 2020-06-23
d3.fsum

### LINKED ISSUE #152: Implement exact summation (fsum)?
Discussed with Kahan summation (#33, #149), fsum is a bit slower but provably "exact" and would be the preferred "exact summation" algorithm for d3-array.

I've implemented it in https://observablehq.com/d/723090bd2ef54f5f ; it does not return a sum but an object that can be used to continue the sum by adding more terms.

- [x] speed test (about 2x slower than d3.sum which is not that bad)
- [x] a valueof accessor
- [x] check if this implementation is useful in d3-geo.
- [x] document

Fil:
exact summation of IEEE 754 numbers
fixes #152 (see also #33)

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ Creates a full precision adder for [IEEE 754](https://en.wikipedia.org/wiki/IEEE
 
 <a name="adder_add" href="#adder_add">#</a> *adder*.<b>add</b>(number)
 
-Adds *number* to the adder’s current value.
+Adds *number* to the adder’s current value and returns the adder.
 
 <a name="adder_valueOf" href="#adder_valueOf">#</a> *adder*.<b>valueOf</b>()
 
--- a/README.md
+++ b/README.md
@@ -130,6 +130,30 @@ Returns an [unbiased estimator of the population variance](http://mathworld.wolf
 
 Returns the standard deviation, defined as the square root of the [bias-corrected variance](#variance), of the given *iterable* of numbers. If the iterable has fewer than two numbers, returns undefined. An optional *accessor* function may be specified, which is equivalent to calling Array.from before computing the standard deviation. This method ignores undefined and NaN values; this is useful for ignoring missing data.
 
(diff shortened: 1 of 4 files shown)


mbostock:
d3.fsum should just return a number. And it should require an iterable as input, as d3.sum does.

If we want something else for iterative adding, we should expose that as a separate interface, such as a d3.adder class. And I think we should simplify that class substantially: it should just have an add(number) method, no iterables, no accessor function. This class would then be used internally by d3.fsum, and if people want the more expressive interface they can use d3.adder instead of d3.fsum.

Fil:
I've fixed the API and made the code more readable.

README.md
```diff
@@ -130,6 +130,29 @@ Returns an [unbiased estimator of the population variance](http://mathworld.wolf
...
+
+Although slower, d3.fsum can replace d3.sum everywhere an exact summation is needed. Uses <a href="#adder">d3.Adder</a>.
+
+<a name="adder" href="#adder">#</a> new d3.<b>Adder</b>()
+
+Creates a full precision adder for [IEEE 754](https://en.wikipedia.org/wiki/IEEE_754) floating point numbers, setting its value to 0.
+
+<a name="adder_add" href="#adder_add">#</a> *adder*.<b>add</b>(number)
+
+Adds *number* to the adder’s current value.
```
mbostock, line 150:
```suggestion
Adds *number* to the adder’s current value and returns the adder.
```

src/fsum.js
```diff
@@ -0,0 +1,60 @@
+// adapted from https://github.com/python/cpython/blob/master/Modules/mathmodule.c#L1444
+export class Adder {
+  constructor() {
+    this.partials = new Float64Array(32).fill(0);
+    this.n = 0;
```
mbostock, line 5: These should probably have underscores to indicate that they are internal variables.

```suggestion
    this._partials = new Float64Array(32).fill(0);
    this._n = 0;
```

src/fsum.js
```diff
@@ -0,0 +1,60 @@
+// adapted from https://github.com/python/cpython/blob/master/Modules/mathmodule.c#L1444
```
mbostock, line 1:
```suggestion
// https://github.com/python/cpython/blob/a74eea238f5baba15797e2e8b570d153bc8690a7/Modules/mathmodule.c#L1423
```

Issue #138, opened 2020-03-06
d3.bisectCenter?


mbostock:
It’d be nice to have something that returned the closest datum to the specified value. I’m often doing something like this:

```js
bisect = {
  const bisectDate = d3.bisector(d => d.date).left;
  return (data, date) => {
    const i = bisectDate(data, date, 1);
    const a = data[i - 1], b = data[i];
    return date - a.date > b.date - date ? b : a;
  };
}
```

It’d be nicer to say

```js
bisect = d3.bisector(d => d.date).center
```

(Obviously, this would only work for quantitative values; ordinal values aren’t sufficient.)

mbostock:
Here’s another example:

https://observablehq.com/@d3/multi-line-chart

Fil:
the (d) accessor / (d, x) comparator thing was a bit tricky, but I think #156 works as expected.

PR #156, opened 2020-06-24
bisector.center

Fil:
fixes https://github.com/d3/d3-array/issues/138


Fil:
And the demo https://observablehq.com/d/8581f0e7e07ee62b

mbostock:
I found it surprising that *bisector*.center returns values rather than indexes, unlike *bisector*.left and *bisector*.right. So I’ve also rewritten the code to return indexes. And I noticed that it wasn’t consistent with *bisector*.left in the ordinal case, so I’ve also added some more tests.

PR #158, opened 2020-06-25
generalize transpose to [[]], [{}], {[]} and {{}}

### LINKED ISSUE #48: transpose(object)?
We have [transpose(*matrix*)](https://github.com/d3/d3-array/blob/master/README.md#transpose) which transposes the rows and columns of a *matrix* (an array of arrays). But you might want similar functionality to transpose an object whose values are arrays into an array whose elements are objects.

```js
function transpose(object) {
  var m = 0;
  for (var k in object) m = Math.max(m, object[k].length);
  for (var i = -1, transpose = new Array(m); ++i < m;) {
    var o = transpose[i] = {};
    for (var k in object) {
      o[k] = object[k][i];
    }
  }
  return transpose;  
}
```

Where:

```js
transpose({
  year: [2001, 2002],
  value: [1, 2]
})
```

Returns:

```json
[
  {"year": 2001, value: 1},
  {"year": 2002, value: 2},
]
```

Should we try to overload d3.transpose to do both?

If so, how? The above implementation sort-of works for array-of-array input such as `transpose([[0, 1, 2], [3, 4, 5]])`, but returns an array of objects rather than an array of arrays. You could use Array.isArray to test whether the input is an Array and branch the behavior accordingly.

If not, what name should this new method have? transposeObject?

Also, if *transpose*(*object*) given an object whose values are arrays returns an array of objects, then *transpose*(*array*) given an array whose elements are objects should return an object whose values are arrays.

Maybe this is too magical.

Fil:
closes #48

- [ ] I'm not sure if Object.keys/Object.entries are allowed yet, easy to change or polyfill.
- [ ] This is about 3x slower in the case of a regular matrix; is it worth keeping the current code and use it if matrix.length > 0 && matrix[0].length > 0 ? (https://observablehq.com/d/dff751d1da207b30)

### DIFF
--- a/README.md
+++ b/README.md
@@ -526,7 +526,7 @@ d3.range(49).map(function(d) { return d / 49; }); // GOOD: returns 49 elements.
 
 <a name="transpose" href="#transpose">#</a> d3.<b>transpose</b>(<i>matrix</i>) · [Source](https://github.com/d3/d3-array/blob/master/src/transpose.js), [Examples](https://observablehq.com/@d3/d3-transpose)
 
-Uses the [zip](#zip) operator as a two-dimensional [matrix transpose](http://en.wikipedia.org/wiki/Transpose).
+A two-dimensional [matrix transpose](http://en.wikipedia.org/wiki/Transpose). Accepts matrices as arrays of arrays [[]], arrays of objects [{}], objects of arrays {[]} and objects of objects {{}}.
 
 <a name="zip" href="#zip">#</a> d3.<b>zip</b>(<i>arrays…</i>) · [Source](https://github.com/d3/d3-array/blob/master/src/zip.js), [Examples](https://observablehq.com/@d3/d3-transpose)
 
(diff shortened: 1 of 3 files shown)


mbostock:
> The weird part might be `length ? [] : {};` because I don't know how to distinguish [] and {} otherwise?

That’s exactly it. There are a variety of other ways to test for arrays (or iterables), and it’s not clear to me that the truthy length property is what we want. My thought was that if we don’t try to handle all of these cases in one method then we can more precisely define what the behavior should be. The current documentation doesn’t really say what will happen, and I think it would be cumbersome to define it.

Fil:
d3.permute is agnostic to [] {}, but its keys are passed explicitly
https://github.com/d3/d3-array/blob/master/src/permute.js

Fil:
Would it be more acceptable to test for `x instanceof Array` instead?

In any case, whether we instantiate the row or column as a [] or a {} is a bit cosmetic, since arrays can have keys other than integers, and vice versa.

Also, if these are iterables we should create an Array.from them?

Fil:
closing due to inactivity

PR #159, opened 2020-06-25
quantileIndex, medianIndex

### LINKED ISSUE #140: d3.medianIndex / d3.quantileIndex?
I suggest to also have a function d3.medianIndex(). E.g. it could return an array 

- with one element: the index of the corresponding element - in case the array.length is uneven
- with two elements: the indices of the elements from which the mean was taken - in case the array.length is even

Fil:
closes #140


Fil:
@EE2dev please test :-)

EE2dev:
Here is a test observable:
https://observablehq.com/d/254038e0365760ad

It seems for an array with an 99 (odd number of elements) the value1 and index1 should be the same as value0 and index0 but it isnt. Or am I missing something?

Fil:
Thanks so much for checking this! In the notebook's function I hadn't paid enough attention to the case where the quantile falls exactly on on index; I always returned [i,i+1] in that case. Note that this PR is about the second approach (which I called "ex post" in the notebook).

EE2dev:
Thanks for the clarification. I agree about the downside/ overhead having two indices.. I am fine with your suggestion.

Fil:
I pushed a new version that returns the left-side index

mbostock (approved):
In 11fa241, I made some changes:

* I changed the implementation of quantileIndex to match [my previous suggestion](https://github.com/d3/d3-array/pull/159#issuecomment-678790995), though I needed to use greatest instead of max. The previous approach was doing a linear scan after computing the quantile value, but we can compute the quantile index directly, so I think it’s preferable to do that. Performance-wise I expect them to be similar since both require making two copies of the input array.
* We need to apply numeric coercion even if valueof is undefined; the behavior of quantileIndex should match quantile.
* (And even so, we should test `valueof` for strictly undefined rather than falsey to match the behavior of default arguments in JavaScript.)
* Tests should use strict equality instead of loose equality wherever possible.

FWIW, here is another alternative implementation that I considered, but I prefer the one now in this PR.

```js
function quantileIndex(values, p, valueof) {
  values = Float64Array.from(numbers(values, valueof));
  if (!(n = values.length)) return;
  if ((p = +p) <= 0 || n < 2) return minIndex(values);
  if (p >= 1) return maxIndex(values);
  var n,
      m = Math.floor((n - 1) * p),
      q = max(quickselect(values.slice(), m).subarray(0, m + 1)),
      i = -1;
  while (++i < n) {
    if (values[i] === q) {
      return i;
    }
  }
}
```

PR #166, opened 2020-08-26
Add bisectCenter.

mbostock:
Looks like we forgot to add the identity bisector for *bisector*.center. #138

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,12 @@ Returns the insertion point for *x* in *array* to maintain sorted order. The arg
 
 Similar to [bisectLeft](#bisectLeft), but returns an insertion point which comes after (to the right of) any existing entries of *x* in *array*. The returned insertion point *i* partitions the *array* into two halves so that all *v* <= *x* for *v* in *array*.slice(*lo*, *i*) for the left side and all *v* > *x* for *v* in *array*.slice(*i*, *hi*) for the right side.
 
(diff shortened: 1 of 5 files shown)


mbostock:
Related question: because *hi* defaults to *array*.length, this doesn’t do what I want (return the index of the closest value) when *value* is greater than *array*[*array*.length - 1]:

```js
const i = d3.bisectCenter(array, value);
```

Instead I need to say:

```js
const i = d3.bisectCente(array, value, 0, array.length - 1);
```

I wonder if this is a mistake and we should instead default *hi* to *array*.length - 1 in the case of *bisector*.center?

mbostock:
Okay, given that the documented behavior is “finds the index of the closest value”, this is clearly a bug. I’ve fixed the bug and added a test.

Fil (approved):
By coercing the values to numbers we now have:

> d3.bisectCenter(["A", "B", "C", "D"], "C"); // 0

where the previous version was giving 2.

This is not what I expect from the documentation:
> This expects that the bisector’s associated accessor returns a quantitative value, or that the bisector’s associated comparator returns a signed distance; **otherwise, this method is equivalent to *bisector*.left.**

> d3.bisectLeft(["A", "B", "C", "D"], "C"); // 2
> d3.bisector(d3.ascending).center(["A", "B", "C", "D"], "C"); // 2

If we keep the code as is (which makes sense), then the documentation for bisectCenter should be expanded a bit 
> [*bisector*.center](#bisector_center) with a numeric substraction comparator

(or something in better English)

Fil:
https://observablehq.com/@d3/d3-bisect will need a small update

mbostock:
I’ve updated the documentation to say that d3.bisectCenter requires an array of numbers (as opposed to ordinal values).

Fil (approved):

PR #168, opened 2020-08-27
Add d3.shuffler.

### LINKED ISSUE #167: Seeded shuffle.
There’s no easy way to pass in a seeded random number generator to d3.shuffle. Let’s make it possible.

mbostock:
Testing the nondeterministic version isn’t feasible because Math.random is captured on load (so monkey-patching no longer works, but there was no documented guarantee that would work anyway). So the nondeterministic test checks that the data isn’t sorted which should only fail 1 out of 10! = 3,628,800 times.

fixes #167


Fil (approved):
all good, but I don't see a non-deterministic test?

Fil:
Here's a notebook showing how we would use it on d3js.org's homepage’s “hexamples”.
https://observablehq.com/d/0a9059b1bc661e6b

Creating a random seed in [0,1) seems unnecessarily hard.

mbostock:
Opened https://github.com/d3/d3-random/pull/37.

PR #170, opened 2020-09-12
Add basic array methods for iterables.

mbostock:
What do you think about adding the equivalents of basic array methods (*array*.filter, *array*.sort, *etc.*), but for iterables? Yes, these are all roughly equivalent to converting to an array with Array.from first, but I think it’s convenient (and hopefully faster) to have dedicated methods.

For example:

```js
const set = new Set([1, 2, 3]);
every(set, d => d > 0); // true
```


mbostock:
Another possibility to consider is whether we want to try to preserve the input type rather than always returning an Array. I think we could do that for typed arrays, Set, and Map, if we wanted. But I worry that it might be a little confusing.

Also, we should note in the documentation that these methods behave slightly different than the array methods with respect to sparse arrays: because the basic array methods skip missing entries, but array iterators yield undefined values.

Fil:
I've often been wanting to type `(iterator).filter(…)` as a chained method. I like the fact that the verb comes after the noun/subject. filter(iterator, …) is not as good, but I suppose it's only a question of subclassing Set or Map? Would d3.group use an "extended" Map?

mbostock:
I’d prefer to use the standards than to create new classes to avoid needing to convert to or from.

The proposed [pipeline operator](https://github.com/tc39/proposal-pipeline-operator) will let you use standard functions with nice chaining syntax. There’s even a partial application proposal but you can also wrap with an arrow function.

mbostock:
Added tests; this should be ready for review.

I thought more about type preservation but I like the conceptual simplicity of always returning an array. You can use *array*.filter or *array*.map if you want to preserve typed array, or pass to new Set or new Map for collections.

mbostock:
I also thought about lodash [_.matches](https://lodash.com/docs/4.17.15#matches) and [_.property](https://lodash.com/docs/4.17.15#property) style shorthand, but I feel that now that JavaScript supports arrow functions,

```js
filter(objects, matches({a: 4}))
map(objects, property("a.b"))
```

is just a more cryptic form of

```js
filter(objects, o => o.a === 4)
map(objects, o => o.a.b)
```

And especially with things like optional chaining now in the language there’s no reason to create a string-based alternative to native syntax.

src/every.js
```diff
@@ -0,0 +1,10 @@
+export default function every(values, test) {
+  if (typeof test !== "function") throw new TypeError("test is not a function");
+  let index = -1;
```
Fil, line 3: out of curiosity why start at -1 with ++index rather than 0 with index++ ? would save a byte :)

test/every-test.js
```diff
@@ -0,0 +1,53 @@
...
+tape("every(values, test) enforces that values is iterable", (test) => {
+  test.throws(() => d3.every({}, () => true), TypeError);
+});
+
+tape("every(values, test) passes test (value, index, values)", (test) => {
+  const calls = [];
+  const values = new Set([5, 4, 3, 2, 1]);
+  d3.every(values, function() { return calls.push([this, ...arguments]); });
+  test.deepEqual(calls, [
+    [global, 5, 0, values],
```
Fil, line 32: why are we testing this === global?

Fil (approved):

src/every.js
mbostock: Out of habit, mostly. At one point I was told it was faster…

test/every-test.js
mbostock: Because the passed in function should not be called with any particular *this* context.

PR #172, opened 2020-09-18
Add permutations generator

### LINKED ISSUE #171: generating permutations
Would you be open to the addition of a generator function to generate all permutations of an array?

e.g. I have one I ported from [Collections2.java · google/guava](https://github.com/google/guava/blob/f4b3f611c4e49ecaded58dcb49262f55e56a3322/guava/src/com/google/common/collect/Collections2.java#L622-L683) that I could contribute. It is an implementation of the Plain Changes algorithm for permutations generation described in Knuth's "The Art of Computer Programming", Volume 4, Chapter 7, Section 7.2.1.2.

mfulton26:
Closes #171.

### DIFF
--- a/README.md
+++ b/README.md
@@ -532,9 +532,31 @@ d3.pairs([1, 2, 3, 4], (a, b) => b - a); // returns [1, 1, 1];
 
 If the specified iterable has fewer than two elements, returns the empty array.
 
-<a name="permute" href="#permute">#</a> d3.<b>permute</b>(<i>source</i>, <i>keys</i>) · [Source](https://github.com/d3/d3-array/blob/master/src/permute.js), [Examples](https://observablehq.com/@d3/d3-permute)
+<a name="permute" href="#permute">#</a> d3.<b>permute</b>(<i>source</i>[, <i>keys</i>]) · [Source](https://github.com/d3/d3-array/blob/master/src/permute.js), [Examples](https://observablehq.com/@d3/d3-permute)
 
-Returns a permutation of the specified *source* object (or array) using the specified iterable of *keys*. The returned array contains the corresponding property of the source object for each key in *keys*, in order. For example:
+Returns a generator of permutations or a single permutation of the specified *source* object (or array).
+
+If *keys* is omitted then a [generator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator) is returned which, when iterated, generates all permutations of the input. For example:
+
+```js
+const permutations = permute(["a", "b", "c"]);
+permutations.next().value; // returns ["a", "b", "c"]
+permutations.next().value; // returns ["a", "c", "b"]
+permutations.next().value; // returns ["c", "a", "b"]
+permutations.next().value; // returns ["c", "b", "a"]
+permutations.next().value; // returns ["b", "c", "a"]
+permutations.next().value; // returns ["b", "a", "c"]
+```
+
+Generators are iterable and can be used in [`for…of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of):
+
+```js
+for (const permutation of permute(["a", "b", "c"])) {
+  // code looping over each permutation
+}
+```
+
+If *keys* is specified then a single permutation is returned. The returned array contains the corresponding property of the source object for each key in *keys*, in order. For example:
 
 ```js
 permute(["a", "b", "c"], [1, 2, 0]); // returns ["b", "c", "a"]
--- a/src/permute.js
+++ b/src/permute.js
@@ -1,3 +1,45 @@
-export default function(source, keys) {
-  return Array.from(keys, key => source[key]);
+export default function (source, keys) {
+  if (keys === undefined) {
+    return permute(source);
+  } else {
+    return Array.from(keys, (key) => source[key]);
+  }
+}
+
+// Based on https://github.com/google/guava/blob/f4b3f611c4e49ecaded58dcb49262f55e56a3322/guava/src/com/google/common/collect/Collections2.java#L622-L683
+// Apache-2.0 License, Copyright (C) 2008 The Guava Authors.
+export function* permute(source) {
+  if (source.length <= 1) {
+    yield source.slice();
+  } else {
+    source = source.slice();
+    const c = Array(source.length).fill(0);
+    const o = Array(source.length).fill(1);
+    let j = Infinity;
+    while (j > 0) {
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -532,9 +532,31 @@ d3.pairs([1, 2, 3, 4], (a, b) => b - a); // returns [1, 1, 1];
 
 If the specified iterable has fewer than two elements, returns the empty array.
 
-<a name="permute" href="#permute">#</a> d3.<b>permute</b>(<i>source</i>, <i>keys</i>) · [Source](https://github.com/d3/d3-array/blob/master/src/permute.js), [Examples](https://observablehq.com/@d3/d3-permute)
+<a name="permute" href="#permute">#</a> d3.<b>permute</b>(<i>source</i>[, <i>keys</i>]) · [Source](https://github.com/d3/d3-array/blob/master/src/permute.js), [Examples](https://observablehq.com/@d3/d3-permute)
 
-Returns a permutation of the specified *source* object (or array) using the specified iterable of *keys*. The returned array contains the corresponding property of the source object for each key in *keys*, in order. For example:
```
curran, line 537: This is really good. No detail of the original documentation was lost here. Nice work!

README.md
```diff
@@ -532,9 +532,31 @@ d3.pairs([1, 2, 3, 4], (a, b) => b - a); // returns [1, 1, 1];
...
+Returns a generator of permutations or a single permutation of the specified *source* object (or array).
```
curran, line 537:
```suggestion
Returns a single permutation (if *keys* is specified) or a generator of all possible permutations (if *keys* is omitted) of the specified *source* object (or array).
```

This language "Returns a generator of permutations or a single permutation of the specified *source* object (or array)." is ambiguous in terms of "a generator of" possibly being associated with "a single permutation". Meaning, one could interpret this sentence as saying that it returns a generator with a single permutation. To solve this, I suggest to reverse the order, so the generator concept is only introduced in association with the notion of "all possible permutations" (another level of detail I've suggested here). I also suggest that this first line of documentation make it clear the conditions with which either case will happen, so I've included those in parentheses.

It might make sense to also flip the more detailed content to match this ordering, so first

>If *keys* is specified then...

then later

>If *keys* is omitted then...

src/permute.js
```diff
@@ -1,3 +1,45 @@
...
-  return Array.from(keys, key => source[key]);
+export default function (source, keys) {
+  if (keys === undefined) {
+    return permute(source);
+  } else {
+    return Array.from(keys, (key) => source[key]);
+  }
+}
+
+// Based on https://github.com/google/guava/blob/f4b3f611c4e49ecaded58dcb49262f55e56a3322/guava/src/com/google/common/collect/Collections2.java#L622-L683
```
curran, line 9: Maybe this link can be moved into the docs?

src/permute.js
```diff
@@ -1,3 +1,45 @@
...
+// Apache-2.0 License, Copyright (C) 2008 The Guava Authors.
```
curran, line 10: Confusion on licensing here. Not sure how to address, but this comment is super confusing. Does it override the licence for the overall library? Is this code dual licensed with both the library license and this licence here in the comment? I suggest to remove this comment, and somehow come to terms with the fact that you are re-licensing this code/algorithm by submitting this PR.

The original license does not apply to a JavaScript port, I don't think.

src/permute.js
curran: Something like.

This implementation was inspired by the [permute algorithm in Guava Collections](https://github.com/google/guava/blob/f4b3f611c4e49ecaded58dcb49262f55e56a3322/guava/src/com/google/common/collect/Collections2.java#L622-L683).

test/permute-test.js
```diff
@@ -48,3 +146,12 @@ tape("permute(…) can take a typed array as the source", (test) => {
 tape("permute(…) can take an iterable as the keys", (test) => {
   test.deepEqual(d3.permute({foo: 1, bar: 2}, new Set(["bar", "foo"])), [2, 1]);
 });
+
+function count(values) {
+  let count = 0;
+  // eslint-disable-next-line no-unused-vars
```
curran, line 152: There must be a way to do this without unused vars?

test/permute-test.js
```diff
@@ -1,6 +1,104 @@
 const tape = require("tape-await");
 const d3 = require("../");
 
+//#region Based on https://github.com/google/guava/blob/f4b3f611c4e49ecaded58dcb49262f55e56a3322/guava-tests/test/com/google/common/collect/Collections2Test.java
+
+tape("permute(…) permutes zero values", (test) => {
```
curran, line 6: Very nice tests!

curran (changes requested):
Looks great overall! I did a thorough review and identified some minor issues, and provided suggestions for each.

mbostock:
We don’t want d3.permute to have a radically different semantics if only a single argument is passed. That’s very confusing. So, this needs a new name. I propose d3.permutations.

I’m not sure I see this function reaching the bar of usefulness to merit inclusion in this library. Do you have some practical, applied examples?

Given that this is derived from third-party code rather than an original contribution, this would require including the Google Collections Apache license. Also, that algorithm is a little complicated, and is it worth the complexity? I feel it would be simpler and easier to write a fresh implementation of the simpler [Heap’s algorithm](https://en.wikipedia.org/wiki/Heap's_algorithm) such as

```js
function* permutations(array) {
  array = Array.from(array); // defensive copy and convert to array
  let i = 0, n = array.length, index = new Uint32Array(n).fill(0);
  yield array.slice(); // defensive copy
  while (i < n) {
    let j = index[i];
    if (j < i) {
      if ((i & 1) === 0) j = 0;
      let t = array[j];
      array[j] = array[i];
      array[i] = t;
      yield array.slice(); // defensive copy
      ++index[i];
      i = 0;
    } else {
      index[i] = 0;
      ++i;
    }
  }
}
```

I guess if it’s that small, there’s not much harm in including it… but I’d still appreciate some applied examples.

Another consideration is whether we want to implement an optional *r* argument, and a d3.combinations method at the same time. For reference, here’s [a port of itertools](https://observablehq.com/@jrus/itertools) by Jacob Rus, and the [Python itertools API reference](https://docs.python.org/3/library/itertools.html) for [itertools.permutations](https://docs.python.org/3/library/itertools.html#permutations) and [itertools.combinations](https://docs.python.org/3/library/itertools.html#combinations). Of course if we’re going to base on implementation on Python’s itertools then we’ll likewise need to add a copy of the [PSF license](https://docs.python.org/3/license.html).

Issue #46, opened 2016-12-31
How to create histogram with fixed bin width


netzwerg:
I am trying to create a histogram with a specific number of bins which should all have the same width (i.e. the domain should be uniformly divided):

I started by using `x.ticks()` (Option 1):

```javascript
const binCount = 5;
const data = [1, 2, 3, 4, 4.7];
const [min, max] = d3.extent(data);
const x = d3.scaleLinear().domain([min, max]);

// Option 1: Tick-based thresholds

const histogram1 = d3.histogram()
  .domain(x.domain())
  .thresholds(x.ticks(binCount));
const bins1 = histogram1(data);

console.log("Option 1 bin widths: " + bins1.map(b => (b.x1 - b.x0)));
// 1,1,1,0.7000000000000002
```
The last bin is narrower than all other bins (to be expected based on how `ticks` works).

For `histogram.thresholds([count])`, the docs state that:

> If a count is specified instead of an array of thresholds, then the domain will be uniformly divided into approximately count bins

This is not what I observe (Option 2):

```javascript
// Option 2: Count-based thresholds

const histogram2 = d3.histogram()
  .domain(x.domain())
  .thresholds(binCount);

const bins2 = histogram2(data);

console.log("Option 2 bin widths: " + bins2.map(b => (b.x1 - b.x0)));
// 1,1,1,0.7000000000000002
```

**Q: What's the proper invocation of `histogram.thresholds([count])`?**

I currently use a manual array of thresholds (Option 3):

```javascript
// Option 3: Range-based thresholds

const thresholds = d3.range(min, max, (max - min) / binCount);
const histogram3 = d3.histogram()
  .domain(x.domain())
  .thresholds(thresholds);

const bins3 = histogram3(data);

console.log("Option 3 bin widths: " + bins3.map(b => (b.x1 - b.x0)));
// 0.74,0.74,0.7399999999999998,0.7400000000000002,0.7400000000000002
```

This works, but seems overly complex for such a simple use case...

mbostock:
Regarding the first or last bin not being the same width as the others: that is because given *n* thresholds, *n* + 1 bins will be produced. The first bin (`bins[0]`) contains any value less than the first threshold (`thresholds[0]`); the last bin (`bins[thresholds.length]`) contains any value greater than or equal to the last threshold (`thresholds[thresholds.length - 1]`). 

[...]

PR #174, opened 2020-09-22
A nice default bin domain.

mbostock:
This changes the default behavior of d3.bin to nice the domain such that the bins are uniform width. The previous behavior often resulted in thinner first and last bins, which was confusing. #46 To make this easier, I’ve also added d3.nice which implements iterative nicing strategy now used by d3-scale (and d3-scale can be upgraded to use d3.nice in the future, if desired).

### DIFF
--- a/README.md
+++ b/README.md
@@ -574,11 +574,15 @@ Ticks are inclusive in the sense that they may include the specified *start* and
 
 <a name="tickIncrement" href="#tickIncrement">#</a> d3.<b>tickIncrement</b>(<i>start</i>, <i>stop</i>, <i>count</i>) · [Source](https://github.com/d3/d3-array/blob/master/src/ticks.js), [Examples](https://observablehq.com/@d3/d3-ticks)
 
-Like [d3.tickStep](#tickStep), except requires that *start* is always less than or equal to *step*, and if the tick step for the given *start*, *stop* and *count* would be less than one, returns the negative inverse tick step instead. This method is always guaranteed to return an integer, and is used by [d3.ticks](#ticks) to guarantee that the returned tick values are represented as precisely as possible in IEEE 754 floating point.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -580,6 +580,10 @@ Like [d3.tickStep](#tickStep), except requires that *start* is always less than
 
 Returns the difference between adjacent tick values if the same arguments were passed to [d3.ticks](#ticks): a nicely-rounded value that is a power of ten multiplied by 1, 2 or 5. Note that due to the limited precision of IEEE 754 floating point, the returned value may not be exact decimals; use [d3-format](https://github.com/d3/d3-format) to format numbers for human consumption.
 
+<a name="nice" href="#nice">#</a> d3.<b>nice</b>(<i>start</i>, <i>stop</i>, <i>count</i>)
+
+Returns a new interval [*niceStart*, *niceStop*] covering the given interval [*start*, *stop*] and where *niceStart* and *niceStop* are guaranteed to align with the corresponding [tick step](#tickStep). Like [d3.tickIncrement](#tickIncrement), this requires that *start* is less than or equal to *step*.
```
Fil, line 585: I realize that *step* is mentioned but not really defined? (except maybe by looking very hard at the definition of tickStep)

Fil (approved):
a very welcome change!

will need to update https://observablehq.com/@d3/d3-bin##last%20bucket

README.md
mbostock: That’s a typo. I meant *stop*.

Fil: ahha, how come I didn't see it? Anyway, I searched for *step* and found that it is also used (and not defined) at https://github.com/d3/d3-array#tickIncrement — in fact probably the same typo.

mbostock: Good catch! Fixed that typo, too.

PR #175, opened 2020-09-22
Add basic set methods.

### LINKED ISSUE #173: Set operations: intersection, union, difference, etc.
https://docs.python.org/3/library/stdtypes.html#set

mbostock:
Fixes #173. Needs a README update.


Fil:
Two small remarks:

1. It's very nice that it can consume strings (as sets of characters).
>  d3.difference("Hello", "world") // Set{H,e}

2. I'm a bit worried that the word "difference" would be used for set difference (or "relative complement" if I read WP correctly),  since it's a word with [many meanings](https://en.wikipedia.org/wiki/Difference#Science_and_mathematics). But d3.setDifference or d3.relativeComplement would be heavy on the stomach.

mbostock:
There’s precedent for “difference” (meaning set difference) in [lodash](https://lodash.com/docs/4.17.15#difference). It’d be nicer if Set supported this natively…

PR #177, opened 2020-09-26
Fix zero-width last bin.

mbostock:
This follows #174 to fix an edge case where the input data’s maximum value is coincident with the niced upper bound of the domain. When this occurs, we extend the niced upper bound by one additional tick step to again ensure uniform bin widths.

Ref. https://observablehq.com/d/a9da34f2bbf326b0

### DIFF
--- a/src/bin.js
+++ b/src/bin.js
@@ -4,7 +4,7 @@ import constant from "./constant.js";
 import extent from "./extent.js";
 import identity from "./identity.js";
 import nice from "./nice.js";
-import ticks from "./ticks.js";
+import ticks, {tickIncrement} from "./ticks.js";
 import sturges from "./threshold/sturges.js";
 
 export default function() {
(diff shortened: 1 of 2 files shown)


src/bin.js
```diff
@@ -29,13 +29,32 @@ export default function() {
...
+
+      // If the last threshold is coincident with the domain’s upper bound, the
+      // last bin will be zero-width. If the default domain is used, and this
+      // last threshold is coincident with the maximum input value, we can
+      // extend the niced upper bound by one tick to ensure uniform bin widths;
+      // otherwise, we simply remove the last threshold.
+      if (tz[tz.length - 1] === x1) {
+        if (x1 === max && domain === extent) {
+          const step = tickIncrement(x0, x1, tn);
+          if (isFinite(step)) {
```
Fil, line 47: Not clear to me when tickIncrement would not be finite?

I looked up the documentation says “this method is always guaranteed to return an integer”, a comment that seems slightly misleading since:
- d3.tickIncrement(0, 1, 0) returns Infinity
- d3.tickIncrement(0, 0, 0) returns NaN
- d3.tickIncrement(0, 0, 1) return -Infinity

I haven't been able to have it return 0, a case which is tested L50 below, as well as on https://github.com/d3/d3-array/blob/master/src/ticks.js#L15

Fil (approved):
I added a tangential remark but the code works well.

PR #183, opened 2020-12-01
sort accessor

### LINKED ISSUE #182: d3.sort, but with an accessor (length = 1) instead of a comparator.
It’d be nice if this

```js
d3.sort(array, d => d.value)
```

were shorthand for this

```js
d3.sort(array, (a, b) => d3.ascending(a.value, b.value))
```

though where the accessor is only evaluated once (in case of non-determinism). We already do something similar for d3.bisector so this seems reasonable!

mbostock:
Fixes #182.

### DIFF
--- a/README.md
+++ b/README.md
@@ -672,12 +672,27 @@ d3.reverse(new Set([0, 2, 3, 1])) // [1, 3, 2, 0]
 ```
 
 <a name="sort" href="#sort">#</a> d3.<b>sort</b>(<i>iterable</i>, <i>comparator</i> = d3.ascending) · [Source](https://github.com/d3/d3-array/blob/master/src/sort.js)
+<br><a name="sort" href="#sort">#</a> d3.<b>sort</b>(<i>iterable</i>, <i>accessor</i>)
 
-Returns an array containing the values in the given *iterable* in the sorted order defined by the given *comparator* function. If *comparator* is not specified, it defaults to [d3.ascending](#ascending). Equivalent to [*array*.sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort), except that it does not mutate the given *iterable*, and the comparator defaults to natural order instead of lexicographic order:
+Returns an array containing the values in the given *iterable* in the sorted order defined by the given *comparator* or *accessor* function. If *comparator* is not specified, it defaults to [d3.ascending](#ascending). Equivalent to [*array*.sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort), except that it does not mutate the given *iterable*, and the comparator defaults to natural order instead of lexicographic order:
 
 ```js
 d3.sort(new Set([0, 2, 3, 1])) // [0, 1, 2, 3]
 ```
+
+If an *accessor* (a function that takes a single argument) is specified,
+
+```js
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -672,13 +672,28 @@ d3.reverse(new Set([0, 2, 3, 1])) // [1, 3, 2, 0]
...
+d3.sort(data, d => d.value)
+```
+
+it is equivalent to a *comparator* using [natural order](#ascending):
+
+```js
+d3.sort(data, (a, b) => d3.ascending(a.value, b.value))
+```
+
+The *accessor* is only invoked once per element, and thus may be nondeterministic.
```
Fil, line 695: With this we could almost use the initial index: `d3.sort(A, function(d) { return -arguments[1]; })` works—the only caveat is we can't write `function(d,i)` since it would have a length===2.

src/sort.js
```diff
@@ -1,6 +1,12 @@
 import ascending from "./ascending.js";
+import permute from "./permute.js";
 
-export default function sort(values, comparator = ascending) {
+export default function sort(values, f = ascending) {
   if (typeof values[Symbol.iterator] !== "function") throw new TypeError("values is not iterable");
-  return Array.from(values).sort(comparator);
+  values = Array.from(values);
```
Fil, line 6: Not exactly related to this commit, but aren't we losing something by using plain arrays?

In many cases (such as the optimal transport) I'm sorting a `Float32Array` of indices against values that I'm reading somewhere else.

To take a simple example, I want to sort all the pixels of an image based on their red value. I'd have `indices = Float32Array.from({length: w * h}, (_,i) => i)` and would want to do `sorted = d3.sort(indices, i => imageData.data[4*i])`. I'd rather have `sorted` be a Float32Array (and maybe the intermediate `f`, but I don't think it's generically possible since the accessor might return a string).

Fil (commented):
Yes, that's probably about 99% of usage.

A question is, since it allocates a new array, might it not end up slower or triggering garbage collection, which the comparator version doesn't?

mbostock:
Yes, I expect it might be slower than a simple accessor, but I think it’s more important that it does the right thing even if the accessor is nondeterministic. And the only way to do that is to materialize the values. You can always call *array*.sort or pass a comparator if you care about performance.

README.md
mbostock: Sorry, but I’m not understanding this comment. This:

```js
d3.sort(A, function(d) { return -arguments[1]; })
```

This is an elaborate equivalent to this?

```js
d3.reverse(A)
```

To avoid the length ambiguity we could introduce a separate method, say d3.sortBy, that takes an accessor rather than a comparator. But I think it’s convenient to use the same length test that d3.bisector used. I don’t think you’ll often want to reference the original index in an accessor, and if you did, you could always just do the sorting yourself.

src/sort.js
mbostock: We discussed type preservation of the input array previously in https://github.com/d3/d3-array/pull/170#issuecomment-691697028, and I still think it’s appropriate that these array methods always return plain arrays for generality and simplicity. If you want something fancier, you can call *array*.sort directly.

README.md
Fil: Yes the example was too simple—what I had in mind was more "sorting with reference to an external array of values". But it was more a comment in passing than a specific concern.

PR #185, opened 2021-01-21
Adopt InternMap.

### LINKED ISSUE #184: Adopt InternMap for d3.group et al.?
Should d3.group, d3.rollup, et al. return an [InternMap](https://observablehq.com/@mbostock/internmap) instead of a regular Map? That would allow you to group by date easily. I think we can even consider it backwards-compatible since the behavior is the same for primitives and for objects that do not implement *object*.valueOf (which is most things besides Date and the rarely-used Number, String, and Boolean).

mbostock:
Fixes #184.

Playground: https://observablehq.com/d/d4c5f6ad343866b9

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -50,3 +50,4 @@ export {default as intersection} from "./intersection.js";
 export {default as subset} from "./subset.js";
 export {default as superset} from "./superset.js";
 export {default as union} from "./union.js";
+export {InternMap, InternSet} from "internmap";
--- a/README.md
+++ b/README.md
@@ -51,12 +51,13 @@ var min = d3.min(array);
 * [Iterables](#iterables)
 * [Sets](#sets)
 * [Bins](#bins)
+* [Interning](#interning)
 
 ### Statistics
 
 Methods for computing basic summary statistics.
 
-<a name="min" href="#min">#</a> d3.<b>min</b>(<i>iterable</i>[, <i>accessor</i>]) · [Source](https://github.com/d3/d3-array/blob/master/src/min.js), [Examples](https://observablehq.com/@d3/extent)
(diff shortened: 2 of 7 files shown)


src/index.js
```diff
@@ -50,3 +50,4 @@ export {default as intersection} from "./intersection.js";
 export {default as subset} from "./subset.js";
 export {default as superset} from "./superset.js";
 export {default as union} from "./union.js";
+export {InternMap, InternSet} from "internmap";
```
curran, line 53: Why export these here?

Is the intention for these to end up as `d3.InternMap` and `d3.InternSet` (via [`export * from "d3-array";`](https://github.com/d3/d3/blob/master/index.js#L2))?

mbostock: Yes.

Fil:
Documentation at https://observablehq.com/@d3/d3-group will need an update when this lands. (I've also added a section on d3.index, if you want to proofread it.)

Note that the paragraphs about [key equality](https://observablehq.com/@d3/d3-group#key_equality) are defensive, and we'll need to mention the version number which solves this.

It's worth noting that while InternMap can index on an array, by passing a key function, there is no way to pass a key function to d3.group (and rollup and index). When the key is on the first level it's easy though to upgrade to an InternMap:

```js
const g = new d3.InternMap(
  d3.group(
    [[["foo", "bar"], 1], [["foo", "baz"], 2], [["goo", "bee"], 3]],
    d => d[0]
  ),
  JSON.stringify
);
g.get(["foo", "bar"]); // 1
```

Fil:
I've added documentation in #186.

mbostock:
I thought about exposing the configurable key function through d3.group et al. but that feels like overkill to start. (I mainly think configurable keys are useful for compound keys; compound keys aren’t typically needed with d3.group because it’s already nested, and furthermore JSON.stringify is slow.) If we do need to expose it in the future, it could be something like `d3.group.key(JSON.stringify)(data, d => d.foo)`, or `d3.groupKey` etc.

Thanks for writing docs! I’ll check ’em out now.

PR #193, opened 2021-03-06
fixed d3.range(0, 1, 1/49) returns 50 elements!

likev:
`d3.range(0, 1, step)` is frequently used scenario and I think we should fixed it.


Fil:
See https://github.com/d3/d3-array/pull/150 which introduces d3.linspace

PR #197, opened 2021-03-21
flatRollup

Fil:
Added  flatRollup and tests, then got carried away and added flatGroupSort, which implies we needed a lexicographic comparator for arrays. Maybe that's too much, but it's easy to trim down.


mbostock:
flatRollup seems fine, but not sure I understand the use case for flatGroupSort? That would require array equality to be useful, which JavaScript doesn’t support (even with InternMap).

Fil:
With flatGroupSort(data, d => d.a, d => d.b) the idea is to be able to enumerate the keys in a rolled-up order, in particular to focus on the *n* most populated groups.

When I need to group on two dimensions I'll join them with a joined key: d => `${d.a} ${d.b}`, then recoup my keys with (v => ({a: v[0].a, b: v[0].b, length})) in the reduce part of the rollup function. This is doable but not as comfortable with groupSort, which returns the joined keys as a string, which I then have to split (and be careful with spaces etc).

mbostock:
But you could sort the result of d3.flatGroup, no?

Fil:
Sure! `d3.flatGroupSort(data, v => v.length, d => d.island, d => d.sex)` is the same as `d3.sort(d3.flatGroup(data, d => d.island, d => d.sex), ([, , v]) => v.length).map(([key1, key2]) => [key1, key2])`.
I think I wouldn't even call the last *map*, and keep the values as the last element of the array.

mbostock:
I am onboard with flatRollup but I would not like to ship flatGroupSort and lexicographic (at least not yet) since I’m not sure it’s a justified addition.

PR #200, opened 2021-03-23
Avoid rounding error.

### LINKED ISSUE #199: linear.nice() ticks do not always span the domain
```js
const scale = d3
  .scaleLinear()
  .domain([0.98, 1.13])
  .nice()

scale.domain() // [0.98, 1.14] as expected
scale.ticks() // [0.98, 1, 1.02, 1.04, 1.06, 1.08, 1.1, 1.12]
scale.ticks(17) // [0.98, 0.99, 1, 1.01, 1.02, 1.03, 1.04, 1.05, 1.06, 1.07, 1.08, 1.09, 1.1, 1.11, 1.12, 1.13]
```

I believe this issue is a floating point number precision error in the [d3-array/ticks](https://github.com/d3/d3-array/blob/master/src/ticks.js) 

For this specific example and following the bouncing ball, if [ticks](https://github.com/d3/d3-array/blob/master/src/ticks.js#L5) is called with `(0.98, 1.14, 10)` then we get the following on [line 23](https://github.com/d3/d3-array/blob/master/src/ticks.js#L23)
```js
step = -50; // -> 50
start * step = 0.98 * 50 = 49;
end * step = 1.14 * 50 = 56.999999999999;
```

Naturally calling `floor` unintentionally reduces the `stop` value by 1.

PR [#164](https://github.com/d3/d3-array/pull/164) was added to solve [#162](https://github.com/d3/d3-array/issues/162) however I think it just inverted the issue.

In this scenario the precision is not important so I think a possible solution is to call `toPrecision()` or use `d3-format`?

Happy to submit a PR after discussion.

Note: 
This issue is more an issue for `d3-array` than `d3-scale` but I have submitted it here as this is where the issue surfaced for me. Happy to close this and open one for `d3-array` if required.

Edit:
I note that most of the [tests covering this](https://github.com/d3/d3-scale/blob/master/test/linear-test.js#L385) are using integer domains and a relatively small count (2,3,4). If I call `check([0.98, 1.14], 2/3/4);` then they all pass but not with `count = 10;` (default). 

Should this test suit be expanded a little?

Second Edit:
On second thoughts `toPrecision` wouldn't work due to the unknown size of the `start` on `stop` but possible something as simple as the following:

```js
function round (x) {
  return Math.round(x * 100) / 100
}

if (step > 0) {
  start = Math.ceil(round(start / step));
  stop = Math.floor(round(stop / step));
  ticks = new Array(n = Math.ceil(stop - start + 1));
  while (++i < n) ticks[i] = (start + i) * step;
} else {
  step = -step;
  start = Math.ceil(round(start * step));
  stop = Math.floor(round(stop * step));
  ticks = new Array(n = Math.ceil(stop - start + 1));
  while (++i < n) ticks[i] = (start + i) / step;
}
```

There is possibly a wider issue/effect that I am unaware on but this patch does fix my issue and make all the test pass including that of `d3-array`.

mbostock:
Fixes #199. /cc @jonathonadams

### DIFF
--- a/src/ticks.js
+++ b/src/ticks.js
@@ -15,16 +15,18 @@ export default function(start, stop, count) {
   if ((step = tickIncrement(start, stop, count)) === 0 || !isFinite(step)) return [];
 
   if (step > 0) {
-    start = Math.ceil(start / step);
-    stop = Math.floor(stop / step);
-    ticks = new Array(n = Math.ceil(stop - start + 1));
-    while (++i < n) ticks[i] = (start + i) * step;
+    let r0 = Math.round(start / step), r1 = Math.round(stop / step);
+    if (r0 * step < start) ++r0;
+    if (r1 * step > stop) --r1;
+    ticks = new Array(n = Math.ceil(r1 - r0 + 1));
+    while (++i < n) ticks[i] = (r0 + i) * step;
   } else {
     step = -step;
-    start = Math.ceil(start * step);
-    stop = Math.floor(stop * step);
-    ticks = new Array(n = Math.ceil(stop - start + 1));
-    while (++i < n) ticks[i] = (start + i) / step;
+    let r0 = Math.round(start * step), r1 = Math.round(stop * step);
+    if (r0 / step < start) ++r0;
+    if (r1 / step > stop) --r1;
+    ticks = new Array(n = Math.ceil(r1 - r0 + 1));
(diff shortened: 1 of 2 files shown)


Fil (approved):

jonathonadams:
Thanks for the quick fix.

PR #202, opened 2021-04-26
Adopt type=module

Fil:
follow changes in d3-format:
* type=module
* add exports
* remove zip
* license: ISC
* update dependencies

TODO:
- [x] the tests on *global* fail. (Note: see also https://github.com/d3/d3-selection/pull/276 )
- [x] update README.md
- [x] check LICENSE
- [x] fix json loading
- [x] switch the default branch to *main*


mbostock:
Do the tests on global fail because the code is now in strict mode, so global is undefined? Or is it globalThis?

Fil:
From what I see *global* is OK, but *this* is undefined.

mbostock:
I think since the code is now in strict mode we expect this to be undefined.

Fil:
another (last?) issue, is how to load the barley test data (json file).

what I do here gives an eslint error
https://github.com/d3/d3-array/pull/202/files#diff-d1c9a415ec5b6150109e5d874042dab86178f271a9b3114e383a6058e25dbbdeR4
  5:31  error  Parsing error: Unexpected token import

aulonm:
`deepEqual` from `assert` is deprecated, should use `deepStrictEqual` instead (ref the [documentation](https://nodejs.org/api/assert.html#assert_assert_deepequal_actual_expected_message))

test/.eslintrc.json
```diff
@@ -0,0 +1,15 @@
+{
+  "extends": "eslint:recommended",
+  "parserOptions": {
+    "sourceType": "module",
+    "ecmaVersion": 8
+  },
+  "env": {
+    "es6": true,
```
aulonm, line 8: This needs to be at least version 11 for the `import.meta.url` to work. 

I think we need to decide if we want to go with es6 or jump all the way up to es2020 (which version 11 is part of). It is possible to keep `env es6` and use `ecmaVersion 11`, but not sure if that is the way to go here. (I'd rather just jump all the way up to es2020, or es2021) 

[Ref this discussion here](https://stackoverflow.com/a/65541635)

Fil: I just need to find a way to load the json file that makes everyone happy (ie tests and eslint)

mbostock: Yes, let’s not use require here. Let’s use fs.readFileSync or similar.

aulonm: You can also rename the file to a `.js` instead of `.json`, and inside of the js-file you can do a `default export [...]`. This way its possible to import the js-file into the tests and they will run as usual. Then you wont need fs.readFileSync. 

Not sure what you usual coding practices have been with d3js, so I'll let you guys decide whats best :)

Fil: now let's see if it passes CI :)

PR #204, opened 2021-05-27
Don’t bin nulls

### LINKED ISSUE #203: d3.bin should check for null
Our threshold implementations do already, which is good (via count), but we shouldn’t be passing null values in here:

https://github.com/d3/d3-array/blob/6fda2773d31c2b3377d8b2dbd11a3e334a7cbc9e/src/bin.js#L80-L82

(Undefined and NaN values already fail this test, so we’re good on those.)

mbostock:
Fixes #203.


Fil (approved):
it's a (welcome) breaking change, for d3-array@3 and d3@7

mbostock:
I will do this as part of d3-array@3.

PR #205, opened 2021-05-27
mode

### LINKED ISSUE #201: mode?
Something like:

```js
function mode(values, f) {
  const m = new InternMap();
  let greatest;
  let greatestCount = 0;
  for (const v of Array.from(values, f)) {
    const count = (m.get(v) || 0) + 1;
    if (count > greatestCount) greatestCount = count, greatest = v;
    m.set(v, count);
  }
  return greatest;
}
```

mbostock:
Fixes #201. Like d3.min and d3.max (but not like d3.median), this considers only comparable values and does not coerce them to numbers.


Fil (approved):
Documentation added at https://observablehq.com/@d3/d3-mode
See also #202 (the test will fail with *global*) and #206.

mbostock:
I will do this as part of d3-array@3.

PR #209, opened 2021-06-03
greatest, least, greatestIndex and leastIndex ignore nulls

### LINKED ISSUE #208: least and greatest should ignore nulls
`d3.least([1, null])` returns null, and `d3.greatest([-1, null])` returns null, contrary to the other reducers (such as min, max, median…) which are all skipping nulls.

Interestingly, d3.leastIndex and d3.greatestIndex do not return the same result when using the default and identity accessors (because it follows two different code paths):
```js
d3.leastIndex([1, null]); // 1 [bug]
d3.leastIndex([1, null], d => d); // 0 [correct answer]
```

This issue is similar to #203.

Fil:
closes #208

test the build at https://observablehq.com/@d3/null-and-least-greatest-208

### DIFF
--- a/src/greatest.js
+++ b/src/greatest.js
@@ -7,9 +7,10 @@ export default function greatest(values, compare = ascending) {
     let maxValue;
     for (const element of values) {
       const value = compare(element);
-      if (defined
+      if ((defined
           ? ascending(value, maxValue) > 0
-          : ascending(value, value) === 0) {
+          : ascending(value, value) === 0
+        ) && value !== null) {
         max = element;
         maxValue = value;
         defined = true;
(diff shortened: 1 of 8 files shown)


src/greatest.js
```diff
@@ -7,19 +7,21 @@ export default function greatest(values, compare = ascending) {
     let maxValue;
     for (const element of values) {
       const value = compare(element);
-      if (defined
+      if ((defined
           ? ascending(value, maxValue) > 0
-          : ascending(value, value) === 0) {
+          : ascending(value, value) === 0
+        ) && value !== null) {
```
mbostock, line 13: I think we should do `value != null`, and do this check before invoking ascending (or below, the passed-in comparator).

mbostock (changes requested):

src/greatest.js
Fil: ✅

mbostock:
Superseded by #210.

PR #210, opened 2021-06-03
Don’t consider null comparable.

mbostock:
Alternative to #209.


Fil (approved):
ah yes! you went to the root :)

Issue #147, opened 2020-06-01
d3.cumsum fails to transpile with babel6 + preset-react


TobbeLino:
When trying to transpile "d3-array" with babel 6 + babel-preset-react, babel fails with a syntax error on "cumsum.js"

How to repeat:
Run:
npm install d3-array
npm install babel-cli
npm install babel-preset-react
node node_modules\babel-cli\bin\babel.js --presets babel-preset-react .\node_modules\d3-array\src\cumsum.js

Result:
```
SyntaxError: ./node_modules/d3-array/src/cumsum.js: Only '=' operator can be used for specifying default value. (4:15)
  2 |   var sum = 0, index = 0;
  3 |   return Float64Array.from(values, valueof === undefined
> 4 |     ? v => (sum += +v || 0)
    |                ^
  5 |     : v => (sum += +valueof(v, index++, values) || 0));
  6 | }
```

Omitting "--presets babel-preset-react" and transpiling with e.g. preset-env works:
npm install babel-preset-env
node node_modules\babel-cli\bin\babel.js --presets babel-preset-env .\node_modules\d3-array\src\cumsum.js

Results in:
```
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports.default = cumsum;
function cumsum(values, valueof) {
  var sum = 0,
      index = 0;
  return Float64Array.from(values, valueof === undefined ? function (v) {
    return sum += +v || 0;
  } : function (v) {
    return sum += +valueof(v, index++, values) || 0;
  });
}
```

mbostock:
Sorry, but this isn’t a help forum for Babel. You’ll need to find an alternative forum if you want help using Babel. Good luck!

williaster:
For anyone else that stumbles upon this https://github.com/babel/babel/issues/11038

CallMeProteus:
temporary solution but works by tweaking the input :

 link to original post https://github.com/FormidableLabs/victory-native/issues/653#issuecomment-859398327


function cumsum(values, valueof) {
  let fn = (v) => (sum += +valueof(v, index++, values) || 0);
  if (valueof === undefined)
    fn = (v) => (sum += +v || 0);
  var sum = 0, index = 0;
  return Float64Array.from(values, fn);
}

1rjun:
Can anyone fix this permanently for react js and react native.
Is there any permanent fix available please?

 "react": "17.0.1",
  "react-native": "0.64.0",
  "d3": "^7.0.0",

[...]

PR #214, opened 2021-06-11
Edited the ternary raising bugs

Dilhando:
This ternary is making our app unable to build on React Native (latest).


Fil:
Not sure what the problem is with React. Also, please try and run the unit tests before suggesting a change.

mbostock:
See also #147; this is a bug that should be fixed in Babel. https://github.com/babel/babel/issues/7235 https://github.com/babel/babel/issues/7234

PR #220, opened 2021-07-13
make bins faster by quantization

Fil:
ref. https://github.com/observablehq/plot/issues/454

demo
https://observablehq.com/@d3/faster-bins-454


mbostock (approved):
This looks great. Is there anything else to do here or should we ship it?

PR #223, opened 2021-08-04
use ascendingDefined when sorting by accessor

### LINKED ISSUE #217: sort(A) and sort(A, d => d) are not equivalent

Fil:
use ascendingDefined when we sort with an accessor (and the default identity)

closes #217

### DIFF
--- a/src/sort.js
+++ b/src/sort.js
@@ -4,7 +4,7 @@ import permute from "./permute.js";
 export default function sort(values, ...F) {
   if (typeof values[Symbol.iterator] !== "function") throw new TypeError("values is not iterable");
   values = Array.from(values);
-  let [f = ascending] = F;
+  let [f = ascendingDefined] = F;
   if (f.length === 1 || F.length > 1) {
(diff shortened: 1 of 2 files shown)


mbostock:
How does this compare to #219? My impression is that we should we land #219, but perhaps we could not export ascendingDefined (and drop descendingDefined entirely) if we think there will be confusion with ascending?

Fil:
I probably shouldn't have promoted this branch to a PR, I was just exploring a minimal footprint answer to the d3.sort issue :)

I'll make a PR against #219 to remove descendingDefined and minimally document ascendingDefined.

PR #226, opened 2021-08-14
no anonymous default exports

mbostock:
This is mainly to appease limitations of hot module reload systems (Webpack I think?). I don’t really like having to do this—I don’t know why Webpack et al. can’t treat the default export as an arbitrary name and just work rather than warn, but, whatever… seems easy to fix regardless.

I considered using eslint-plugin-import here, which has a rule for this. But it pulls in many trivial (dev) dependencies and doesn’t seem worth it.

### DIFF
--- a/src/ascending.js
+++ b/src/ascending.js
@@ -1,4 +1,4 @@
-export default function(a, b) {
+export default function ascending(a, b) {
   return a == null || b == null ? NaN
     : a < b ? -1
     : a > b ? 1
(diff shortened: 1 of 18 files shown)


Fil (approved):
I was surprised that the minified version results in being exactly the same (except for the unrelated [change](https://github.com/d3/d3-array/pull/226/files#diff-6d481c69bc82dc9dbec87cdc90a09a5c4b87dc55fec3eaf239981d294d157826) in constant.js).

PR #228, opened 2021-08-17
rank

Fil:
<del>Question: I'm not sure about the API; in particular if we have the idea that the accessor could at some point be specified as a string (like we do in Plot), this would be ambiguous when calling `rank(array, "high")`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ const m = d3.min(array);
 
 Methods for computing basic summary statistics.
 
-<a name="min" href="#min">#</a> d3.<b>min</b>(<i>iterable</i>[, <i>accessor</i>]) · [Source](https://github.com/d3/d3-array/blob/master/src/min.js), [Examples](https://observablehq.com/@d3/d3-extent)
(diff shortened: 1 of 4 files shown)


mbostock:
I think we should drop the _ties_ argument and only implement the “low” strategy for now.

This should support a comparator as well as an accessor like d3.sort and d3.bisector do, probably using compareDefined to protect against non-comparable values. (I suppose this means that d3.rank should accept multiple accessors like d3.sort does, too… but we didn’t extend that to d3.bisector or for that matter d3.quickselect, so maybe let’s punt on that convenience for now and focus on the basics.)

Fil:
I've updated https://observablehq.com/@d3/rank-dev and I think it covers everything (rank, ties, comparator, multiple accessors). Just need to decide on the scope and I'll merge the relevant code and tests.

mbostock:
Would you be okay only supporting the ties = _low_ strategy to start? I would prefer to start there and not support the other strategies.

PR #231, opened 2021-09-18
adopt interning for set operations

mbostock:
This adopts InternSet for d3-array’s set operations, which means you can now use things like d3.union and d3.difference on arrays of dates. As another bonus, you can say `d3.union(values)` instead of `new d3.InternSet(values)`.


mbostock:
The upgrade of internmap is pretty urgent here (the delete operation was broken) so I’m going to go ahead with this. But let me know if you have any objections!

PR #233, opened 2021-09-21
optional log base for ticks functions

### LINKED ISSUE #232: binary ticks increments on linear scale
I would like to propose an added option to allow returning ticks on a binary basis from the `ticks` method.

https://github.com/d3/d3-array/blob/6860b190118e29a08fe9848b97aa1cbe967172f2/src/ticks.js#L5-L35

I imagine this could be an additional parameter like `base` such as...

```js
export default function ticks(start, stop, count, base) {
```

Then this `tickIncrement` logic would be the only change

https://github.com/d3/d3-array/blob/6860b190118e29a08fe9848b97aa1cbe967172f2/src/ticks.js#L37-L44

The changes would be something like...

```diff
-export function tickIncrement(start, stop, count) {
+export function tickIncrement(start, stop, count, base = 10) {
  var step = (stop - start) / Math.max(0, count),
-      power = Math.floor(Math.log(step) / Math.LN10),
+      power = Math.floor(Math.log(step) / Math.log(base) + Number.EPSILON),
-      error = step / Math.pow(10, power);
+      error = step / Math.pow(base, power);
   return power >= 0
-      ? (error >= e10 ? 10 : error >= e5 ? 5 : error >= e2 ? 2 : 1) * Math.pow(10, power)
+      ? (error >= e10 ? 10 : error >= e5 ? 5 : error >= e2 ? 2 : 1) * Math.pow(base, power)
-      : -Math.pow(10, -power) / (error >= e10 ? 10 : error >= e5 ? 5 : error >= e2 ? 2 : 1);
+      : -Math.pow(base, -power) / (error >= e10 ? 10 : error >= e5 ? 5 : error >= e2 ? 2 : 1);
 }
```

Which would look something like...

## Base 2

domain: `[0,1882]`
ticks: `[0,256,512,768,1024,1280,1536,1792]`

Image 2021-09-20 at 3 44 05 PM

## Base 10 (default)

domain: `[0,1882]`
ticks: `[0,200,400,600,800,1000,1200,1400,1600,1800]`

Image 2021-09-20 at 3 44 36 PM

nickofthyme:
Adds optional `base` value to `ticks`, `tickIncrement` and `tickStep` functions. Using default `base` value of `10` results in no functional changes to existing code.

These changes enable rendering linear ticks for [binary](https://mathworld.wolfram.com/BinaryLogarithm.html), [natural](https://mathworld.wolfram.com/NaturalLogarithm.html) and other log bases.

cc: @monfera

closes #232


nickofthyme:
@Fil any update on this?

Fil (changes requested):
I thought I had left a comment, but I now find it was waiting in one of many many open tabs… Sorry! For the documentation on Observable, you can fork the notebook and make a suggestion.

Fil (approved):

mbostock:
My concerns here:

* Too many positional arguments (start, stop, count, base)
* Other implicit dependencies on base 10 (e.g., 2 and 5 are factors of 10, and 10 appears elsewhere in code)
* Tiny loss of precision and performance using Math.log(base) etc. instead of Math.LN10
* In practice will probably only be used for base = 2 and base = e

Overall, I think I would prefer to keep the existing methods as-is and have base-specific variants of d3.ticks and d3.tickIncrement for base = 2 (binary) and base = e (“natural”).

nickofthyme:
That's understandable, I can update the pr to have base-specific function variants.

src/ticks.js
```diff
@@ -2,35 +2,42 @@ const e10 = Math.sqrt(50),
...
     if (i1 * inc < start) ++i1;
     if (i2 * inc > stop) --i2;
   }
-  if (i2 < i1 && 0.5 <= count && count < 2) return tickSpec(start, stop, count * 2);
+  if (i2 < i1 && 0.5 <= count && count < 2) return tickSpec(start, stop, count * 2, log);
   return [i1, i2, inc];
 }
 
-export default function ticks(start, stop, count) {
+export default function ticks(start, stop, count, log) {
```
nickofthyme, line 36: > Too many positional arguments (start, stop, count, base). Overall, I think I would prefer to keep the existing methods as-is and have base-specific variants of d3.ticks and d3.tickIncrement for base = 2 (binary) and base = e (“natural”).

I'm happy to separate these added methods into separate methods. But I'd like to avoid duplication of code. Would you be ok with having the `log` option on a base function then exporting a wrapped function for each log type with just `start`, `stop` and `count` args?

So exporting `ticks`, `ticksNatural` and `ticksBinary`, along with their respective `tickIncrement` functions.

src/ticks.js
nickofthyme, line 1: > Other implicit dependencies on base 10 (e.g., 2 and 5 are factors of 10, and 10 appears elsewhere in code)

Could you point me to the other parts of the code that would need to be addressed?

src/ticks.js
```diff
@@ -2,35 +2,42 @@ const e10 = Math.sqrt(50),
     e5 = Math.sqrt(10),
     e2 = Math.sqrt(2);
 
-function tickSpec(start, stop, count) {
+const logParams = {
+  binary: [Math.log2, 2],
+  natural: [(n) => Math.log1p(n - 1), Math.E],
+  common: [Math.log10, 10],
+}
```
nickofthyme, line 9: > Tiny loss of precision and performance using Math.log(base) etc. instead of Math.LN10

Does this approach work for you to use the respective precise `Math` methods for each allowed log type?

src/ticks.js
```diff
@@ -2,35 +2,42 @@ const e10 = Math.sqrt(50),
     e5 = Math.sqrt(10),
     e2 = Math.sqrt(2);
 
-function tickSpec(start, stop, count) {
+const logParams = {
+  binary: [Math.log2, 2],
+  natural: [(n) => Math.log1p(n - 1), Math.E],
+  common: [Math.log10, 10],
```
nickofthyme, line 8: > In practice will probably only be used for base = 2 and base = e

Limits the log types to just `common`, `natural` and `binary`.

nickofthyme (commented):
@mbostock Sorry for the delay, I finally got around to updating this PR.

I want to follow up on the concerns you mentioned...

PR #235, opened 2021-09-24
Use the commonjs module as the main export

dnjstrom:
The ESM module is still exposed through the dedicated `module` property though for consumers that support it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "src/**/*.js"
   ],
   "module": "src/index.js",
-  "main": "src/index.js",
+  "main": "dist/d3-array.js",
   "jsdelivr": "dist/d3-array.min.js",
   "unpkg": "dist/d3-array.min.js",
   "exports": {


dnjstrom:
This change was prompted due to our jest tests breaking with the following error:

```
/../frontend/node_modules/d3-array/src/index.js:1

SyntaxError: Unexpected token 'export'

    > 1 | import { bisector } from "d3-array";
        | ^
      2 | 

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1726:14)
      at Object.<anonymous> (src/utils.ts:1:1)
```

Our code seem to import the ESM module but Jest doesn't transform esm modules in node_modules, simply expecting commonjs modules from libraries - causing the error. Changing the `main` property to the commonjs export fixes the error but still provide ESM modules for the tools that support them.

mbostock:
Thank you for the suggestion. Per the release notes, D3 now only provides ES modules for Node. CommonJS is no longer supported.

dnjstrom:
Ah, hadn't noticed that - sorry for the noise and thanks for the fast response! 👍

dnjstrom:
For anyone else stumbling across this question, the solution for us was to set the [transformIgnorePatterns](https://jestjs.io/docs/tutorial-react-native#transformignorepatterns-customization) in `jest.config.js` like so:

```json
  transformIgnorePatterns: ["/node_modules/(?!(d3-array|internmap))"],
```

Issue #189, opened 2021-01-25
String accessors?


mbostock:
d3-array currently uses function accessors, e.g.

```js
d3.median(data, d => d.foo)
```

What if a string could be used instead of an accessor for convenience?

```js
d3.median(data, "foo")
```

There’s some precedent for this, e.g. Lodash:

```js
_.map(data, "foo")
```

Here’s where it might be used:

- [ ] d3.min
- [ ] d3.minIndex
- [ ] d3.max
- [ ] d3.maxIndex
- [ ] d3.extent
- [ ] d3.sum
- [ ] d3.mean
- [ ] d3.median
- [ ] d3.cumsum
- [ ] d3.quantile
- [ ] d3.quantileSorted
- [ ] d3.variance
- [ ] d3.deviation
- [ ] d3.fsum
- [ ] d3.least
- [ ] d3.leastIndex
- [ ] d3.greatest
- [ ] d3.greatestIndex
- [ ] d3.bisector
- [ ] d3.group
- [ ] d3.groups
- [ ] d3.index
- [ ] d3.indexes
- [ ] d3.rollup
- [ ] d3.rollups
- [ ] d3.groupSort
- [ ] d3.count
- [ ] d3.every
- [ ] d3.some
- [ ] d3.filter
- [ ] d3.map
- [ ] d3.sort
- [ ] bin.value? (possibly breaking change, but why would you ever use a constant here?)

PR #236, opened 2021-10-01
only implement “low” ties strategy

mbostock:


Fil (approved):
LGTM. If we ever want to add a ties option, and the accessor as a string (#189),
the API would then become
d3.<b>rank</b>(<i>iterable</i>[, <i>accessor</i>[, <i>ties</i>]])
(*i.e.* no ties without an accessor).

PR #237, opened 2021-10-02
allow rank to take a comparator

mbostock:


Fil:
I don't know if it helps, but the comparator is available in https://observablehq.com/@d3/rank-dev, as well as multiple accessors.

Fil (approved):
It works :). And I love the simplicity of the example.

Before I saw the new commits I pushed https://github.com/d3/d3-array/tree/fil/rank-comparator.

It adds:
* accepts multiple accessors too, on par with d3.sort.
* a few unit tests
* moves the documentation of d3.rank below #sort.

let me know if you think any of this should be included.

mbostock:
Thanks! I think it’s nice for d3.rank to be next to d3.quantile since it feels more closely related than d3.sort.

I don’t think I want to commit to multiple accessors at this point. We don’t do that for d3.bisector, for example, and at the least it’s always possible to express as an equivalent comparator now that comparators are supported.

Also, and I’m not sure whether to file a separate issue for this, but using function.length to differentiate an accessor from a comparator feels brittle in retrospect. For example, you’d hope that d3.sort(values, Math.random) would interpret Math.random as an accessor, but since Math.random’s function.length is zero it gets interpreted as a comparator. This confusion bubbles up to Plot e.g. https://github.com/observablehq/plot/pull/564. So, I’m tempted to assume an accessor if function.length !== 2 instead of if function.length === 1; but that change isn’t really backwards-compatible, so we’re a bit stuck.

Fil:
I think it's fair to test function.length < 2 rather than function.length === 1 — is there a use case where a comparator would work with 0 argument? The only one I can think of is if someone did sort on a constant function (to not sort), but then it would also not sort.

PR #246, opened 2022-03-31
fix Scott’s rule

mbostock:
I’m not sure how this happened, but all the literature uses 3.49, but d3-array used 3.5 for some reason.

I also thought about switching to `Math.cbrt(1 / count(values))` but I’m not sure if that’s an improvement.

### DIFF
--- a/src/threshold/scott.js
+++ b/src/threshold/scott.js
@@ -2,5 +2,5 @@ import count from "../count.js";
 import deviation from "../deviation.js";
 
 export default function thresholdScott(values, min, max) {
-  return Math.ceil((max - min) / (3.5 * deviation(values) * Math.pow(count(values), -1 / 3)));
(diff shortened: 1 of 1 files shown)


Fil:
cbrt is more precise than pow(1/3), and faster on Chrome (but on Safari and Firefox the performance is almost identical). None of this matters here :)

Chrome

Fil (approved):

mbostock:
> cbrt is more precise than pow(1/3)

Right, but is cbrt(1 / x) more precise (or faster) than pow(x, -1/3)?

Fil:
I've added those to the test notebook, it doesn't seem to change anything in terms of speed. As for accuracy I don't know what would be relevant. https://observablehq.com/@fil/math-cbrt

PR #248, opened 2022-04-07
fix crash on missing top bin

Fil:
`d3.bin().domain([4, 5])([5])` crashes

the issue is that we have removed the last bin, but then want to assign the largest value to it because floor(x1) === x1
I'm not sure if the fix catches all cases…

issue reported by weiglemc at https://observablehq.com/@weiglemc/demonstrating-issue-with-d3-bins-in-d3v7-4-1-update

### DIFF
--- a/src/bin.js
+++ b/src/bin.js
@@ -84,8 +84,10 @@ export default function bin() {
     if (isFinite(step)) {
       if (step > 0) {
         for (i = 0; i < n; ++i) {
-          if ((x = values[i]) != null && x0 <= x && x <= x1) {
+          if ((x = values[i]) != null && x0 <= x && x < x1) {
             bins[Math.floor((x - x0) / step)].push(data[i]);
+          } else if (x === x1) {
+            bins[bins.length - 1].push(data[i]);
           }
         }
       } else if (step < 0) {
--- a/test/bin-test.js
+++ b/test/bin-test.js
@@ -235,6 +235,10 @@ it("bin(data) assigns floating point values to the correct bins", () => {
   }
 });
 
+it("bin(data) assigns integer values to the correct bins", () => {
+  assert.deepStrictEqual(bin().domain([4, 5])([5]), [box([5], 4, 5)]);
+});
+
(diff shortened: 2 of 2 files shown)


src/bin.js
```diff
@@ -84,8 +84,10 @@ export default function bin() {
     if (isFinite(step)) {
       if (step > 0) {
         for (i = 0; i < n; ++i) {
-          if ((x = values[i]) != null && x0 <= x && x <= x1) {
+          if ((x = values[i]) != null && x0 <= x && x < x1) {
             bins[Math.floor((x - x0) / step)].push(data[i]);
```
mbostock, line 88: Maybe `Math.min(bins.length - 1, Math.floor((x - x0) / step))` here instead?

Fil:
hmmm this is only fixing one of the two original cases… adding a test

Fil:
I don't understand why the test fails with 17 eights, but not with 16 eights!

src/bin.js
```diff
@@ -85,14 +85,14 @@ export default function bin() {
...
-            bins[Math.floor((x - x0) / step)].push(data[i]);
+            bins[Math.min(m, Math.floor((x - x0) / step))].push(data[i]);
           }
         }
       } else if (step < 0) {
         for (i = 0; i < n; ++i) {
           if ((x = values[i]) != null && x0 <= x && x <= x1) {
             const j = Math.floor((x0 - x) * step);
-            bins[j + (tz[j] <= x)].push(data[i]); // handle off-by-one due to rounding
+            bins[Math.min(m, j + (tz[j] <= x))].push(data[i]); // handle off-by-one due to rounding
```
mbostock, line 95: Thinking out loud: if `j` is equal to `m` here (due to the same rounding error), then `tz[j]` will be undefined, so `tj[j] <= x` will be false. I wonder if this Math.min should be above instead, i.e.:

```
const j = Math.min(m, Math.floor((x0 - x) * step));
```

mbostock (approved):

PR #250, opened 2022-04-09
fix comparator bisector

### LINKED ISSUE #249: bisector no longer supports two-argument (object, value) comparator
It looks like 22cdb3f2b1b98593a08907d17c12756404411b1d broke my compare function, which used an array of objects.

I see I can use array of values, but wanted to point out this breaking change.

My compare function was
```
            that.bisectX = d3.bisector(function (d, x) {
                if (that.xascending) {
                    return d.x - x;
                } else {
                    return x - d.x;
                }
            }).left;
```

The breaking change is at https://github.com/d3/d3-array/blob/14efbe0f0f39331df0578e3527e4a0f5d30cc50d/src/bisector.js#L16

mbostock:
Fixes #249. Previously #227.

If the bisector was given an asymmetric comparator that takes (_object_, _value_) as arguments, then the test for whether the search value is orderable was failing because it would pass the comparator (_value_, _value_), which the comparator likely did not expect, and thus would typically return NaN or undefined.

So instead, we now perform the bisection without first testing for orderability, but if the final comparison at the end of bisection returns NaN, we know that the search value was not orderable. In the case that the search value is not orderable, this is obviously slower since we still perform the bisection, but we don’t need to optimize for this case and it’s better to be correct and backwards-compatible.

### DIFF
--- a/src/bisector.js
+++ b/src/bisector.js
@@ -2,35 +2,36 @@ import ascending from "./ascending.js";
 
 export default function bisector(f) {
   let delta = f;
-  let compare1 = f;
-  let compare2 = f;
+  let compare = f;
 
+  // If an accessor is specified, promote it to a comparator.
   if (f.length !== 2) {
     delta = (d, x) => f(d) - x;
-    compare1 = ascending;
-    compare2 = (d, x) => ascending(f(d), x);
+    compare = (d, x) => ascending(f(d), x);
   }
 
   function left(a, x, lo = 0, hi = a.length) {
     if (lo < hi) {
-      if (compare1(x, x) !== 0) return hi;
+      let ohi = hi, c;
       do {
         const mid = (lo + hi) >>> 1;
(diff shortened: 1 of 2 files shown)


mbostock:
Hmm, I just realized that this won’t work because the array could only have non-orderable values in it, and thus the last comparison could return NaN even if the search value is orderable. For example, this now fails:

```js
const values = [null, undefined, NaN];
assert.strictEqual(bisectLeft(values, 1), 0);
assert.strictEqual(bisectLeft(values, 2), 0);
```

If the comparator is specified as a function that takes (_object_, _value_), there’s no way for us to test just an _object_ for orderability (because we can’t construct a synthetic _value_ to test against), just like we can’t test just a _value_ for orderability (because we can’t construct a synthetic _object_ to test against). So… I think we might be stuck here, although we could special case specific comparators that we know to be symmetric. 🤔

PR #253, opened 2022-06-30
Add d3.linspace.

### LINKED ISSUE #90: rangeInclusive?

Fil:
rebases #150 on main
closes #90


mbostock:
I pushed this to the `linspace` branch to update #150 and continue discussion there.

PR #254, opened 2022-07-01
blur2 with stride

Fil:
demo https://observablehq.com/@d3/blur-stride-dev

The API will have to change: having the stride at the end after the optional ry is not the best solution.

There is no need to change blur1 to accept a stride parameter, since it is the same as using blur2 with width=stride, rx=0, ry=r. We can intercept this in the API if we want, for consistency.

### DIFF
--- a/src/blur.js
+++ b/src/blur.js
@@ -11,37 +11,40 @@ export function blur1(values, r) {
   return values;
 }
 
-export function blur2(values, width, rx, ry = rx) {
+export function blur2(values, width, rx, ry = rx, stride = 1) {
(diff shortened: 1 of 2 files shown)


mbostock (changes requested):
This doesn’t feel right to me. It lets you blur RGBA, yes, but you can’t e.g. just blur the R channel, and now blurh and asymmetric with blurv. We need to think through the interface here some more.

mbostock:
Closing in favor of c889f65ac0fbb7cfc5dc35f08bccc607028da91d in #252. Thanks for the help!

Issue #255, opened 2022-07-11
Testing a lib using `d3-array`


tk-o:
## Context 

When I run `jest` tests, I receive the following error:
```
({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){export {default as bisect, bisectRight, bisectLeft, bisectCenter} from "./bisect.js";
                                                                                      ^^^^^^

    SyntaxError: Unexpected token 'export'

      1 | import { useCallback, useEffect } from "react";
      2 |
    > 3 | import { extent, max } from "d3-array";
        | ^
      4 | import { scaleTime, scaleLinear } from "@visx/scale";
      5 |
      6 | import type { PriceDataPoint } from "./types";

      at Runtime.createScriptFromCode (../../node_modules/.pnpm/jest-runtime@28.1.2/node_modules/jest-runtime/build/index.js:1796:14)
      at Object.<anonymous> (src/use-price-graph.hook.ts:3:1)
```

## Solution

I tested this solution locally:
- https://github.com/facebook/jest/issues/12036#issuecomment-1039479569
and it worked immediately.

mbostock:
This library is only available as an ES module on Node (and bundlers). This will need to be fixed in Jest.

PR #256, opened 2022-07-11
fix(jest-integration): add `require` to `exports`

tk-o:
Fixes:
- https://github.com/d3/d3-array/issues/255

Perhaps worth expanding this fix on the remaining libs from the `d3-*` package family.

---

This change follows the advice from:
- https://github.com/facebook/jest/issues/12036#issuecomment-1039479569

Tested locally: the change fixes the problem with `jest` trying to use the `default` path.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
   "jsdelivr": "dist/d3-array.min.js",
   "unpkg": "dist/d3-array.min.js",
   "exports": {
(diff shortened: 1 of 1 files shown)


mbostock:
This library is only available as an ES module on Node (and bundlers). This will need to be fixed in Jest.

PR #258, opened 2022-07-29
Change main entrypoint to dist/index.js

### LINKED ISSUE #255: Testing a lib using `d3-array`

david-bezero:
Provides compatibility with tooling which does not support modules (e.g. Jest). Module exports are already available in "module" for tools which can support them.

Fixes #255

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "src/**/*.js"
   ],
   "module": "src/index.js",
-  "main": "src/index.js",
+  "main": "dist/index.js",
   "jsdelivr": "dist/d3-array.min.js",
   "unpkg": "dist/d3-array.min.js",
   "exports": {


mbostock:
This library is only available as an ES module on Node (and bundlers). This will need to be fixed in Jest.

david-bezero:
@mbostock seems I can't re-open the PR once it has been closed, but here is an alternative approach which might be less objectionable? https://github.com/david-bezero/d3-array/commit/41a69e76390f16d69a9508195a5d2feda393b71a

That re-uses the existing UMD export which you seem to be supporting, but just explicitly calls it out as useable for `require` imports (from testing this works fine in Jest)

As for fixing in Jest: it _might_ be possible to get Jest to fall-back from `require` to `umd` automatically (though I'd argue they shouldn't need to), but Jest certainly won't be supporting ES modules any time soon!

mbostock:
Not that I can think of, sorry. If Jest doesn’t want to support ES modules then I suppose people will be stuck with older versions of D3.

Issue #194, opened 2021-03-11
geometric mean?


Fil:
see the implementation in https://observablehq.com/@fil/d3-gmdn

PR #259, opened 2022-08-15
geometric mean

douglaskariuki:
the geometric mean could also be useful for other types of calculations

### DIFF
--- a/src/gmean.js
+++ b/src/gmean.js
@@ -1,0 +1,34 @@
+export default function gmean(values, valueof) {
+	const r = 64;
+	const K = 2 ** r;
+	const K1 = 2 ** -r;
+	let p = 1; // product
+	let n = 0; // count
+	let s = 1; // sign
+	let k = 0; // track exponent to prevent under/overflows
+	if (valueof === undefined) {
+		for (let value of values) {
+			if (value != null && (value = +value) >= value) {
+					n++;
+					s = Math.sign(value);
+					if (s === 0) return 0;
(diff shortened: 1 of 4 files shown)


Fil:
See #194 and https://observablehq.com/@fil/d3-gmdn

src/gmean.js
```diff
@@ -0,0 +1,34 @@
+export default function gmean(values, valueof) {
+	const r = 64;
```
curran, line 2: Using tabs is inconsistent with the rest of the codebase, which uses spaces.

curran (commented):
I notice that tabs are used in the new file, whereas the rest of the codebase uses spaces.

Also curious, is this implementation derived from any existing work, or inspired from any particular source?

src/gmean.js
douglaskariuki: I changed this commit a few times before pushing and didn't notice the last change was using tabs instead of spaces. Making the necessary changes..

> Using tabs is inconsistent with the rest of the codebase, which uses spaces.

douglaskariuki: > I notice that tabs are used in the new file, whereas the rest of the codebase uses spaces.
> 
> Also curious, is this implementation derived from any existing work, or inspired from any particular source?

Yes, this PR was inspired by d3.mean(), used to calculate the arithmetic mean because I noticed similarities between the two. Will also work on showing differences between the geometric and arithmetic mean calculations and update this PR.

PR #264, opened 2022-12-26
try to generate at least one tick

mbostock:
I noticed that this returns zero ticks:

```js
d3.ticks(1, 364, 1.1)
```

That’s because the chosen tick step is 500, and of course neither 0 or 500 is within [1, 364]. A more egregious case which also returns the empty array for the same reasons is:

```js
d3.ticks(1, 499, 1.5)
```

The fix is simply to double the number of requested ticks recursively when the actual number of ticks is zero. (That’s because tickIncrement does not take into account the alignment of the domain with the tick values.) To avoid any risk of infinite recursion, we only double the count when the count is in the range [0.5, 2); at most two doublings as possible.

This means that this:

```js
d3.ticks(1, 499, 0.5)
```

likewise now returns `[200, 400]` instead of `[]`, but that seems consistent with the documented behavior of returning *count* + 1 = 1.5 values.

### DIFF
--- a/src/ticks.js
+++ b/src/ticks.js
@@ -7,38 +7,38 @@ function tickSpec(start, stop, count) {
     power = Math.floor(Math.log10(step)),
     error = step / Math.pow(10, power),
     factor = error >= e10 ? 10 : error >= e5 ? 5 : error >= e2 ? 2 : 1;
+  let i1, i2, inc;
   if (power < 0) {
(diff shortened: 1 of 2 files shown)


Fil (approved):

mbostock:
Should we do this inside of tickIncrement instead of ticks so that the two are guaranteed to be consistent?

Fil (approved):
nice!

Issue #123, opened 2019-10-04
Float64Array.sort returns undefined in Safari for when length < 2.


mbostock:
In Safari, this returns undefined:

```js
Float64Array.of(0).sort((a, b) => a - b)
```

This causes a crash in d3.quantile when *data*.length < 2.

https://observablehq.com/d/83e268c7b40ffa24

PR #269, opened 2023-02-24
Alter rank to work on Safari 14 when passed a one-element array.

dimfeld:
This is the same issue that was found for `quantile` in #123. Safari 14 returns undefined from `TypedArray.sort` when the array has just 1 element.

I reused `values` to hold the `Uint32Array` instead of creating a new variable, since that's what the similar fix for quantile did in https://github.com/d3/d3-array/commit/35f9d72b5a859b66b5a09773351e80cf11272fa6.


Fil:
Makes sense, but to double check I've installed Safari 14 (from https://www.macupdate.com/app/mac/15675/apple-safari/old-versions) and it doesn't exhibit the bug (Float64Array.of(0).sort((a, b) => a - b) returns [0]).

dimfeld:
We have had a user running Safari 14.1 on Mac hit this in production on our product.

techniq:
@Fil I just installed `14.0.1` on my Mac like you did, and see the same "non-issue"...

image

but using the iOS Simulator with `14.0`, the issue is present...

image

Our error report shows lists Safari `14.1.2` on Max OSX `10.15.7`  🤷‍♂️

Fil (approved):

mbostock (approved):
Looks good. A comment saying this is for a Safari 14 bug would be helpful so we can clean this up in the future when we drop support for old browsers, and so someone doesn’t inadvertently remove this workaround for the sake of code style. (We should have done that in 35f9d72 but oh well!)

dimfeld:
Thanks! Just pushed an update that adds a comment.

Fil:
Merged as https://github.com/d3/d3-array/commit/795283502f6da9e0cd3a70b3a2fce83715f2a740

PR #274, opened 2023-05-09
Update least docs

adebiyial:
This example confused me initially, so I updated the docs to reflect this gotcha.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,13 +212,14 @@ Methods for searching arrays for a specific element.
 Returns the least element of the specified *iterable* according to the specified *comparator* or *accessor*. If the given *iterable* contains no comparable elements (*i.e.*, the comparator returns NaN when comparing each element to itself), returns undefined. If *comparator* is not specified, it defaults to [ascending](#ascending). For example:
 
 ```js
-const array = [{foo: 42}, {foo: 91}];
+const array = [{foo: 91}, {foo: 42}];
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -212,13 +212,14 @@ Methods for searching arrays for a specific element.
 Returns the least element of the specified *iterable* according to the specified *comparator* or *accessor*. If the given *iterable* contains no comparable elements (*i.e.*, the comparator returns NaN when comparing each element to itself), returns undefined. If *comparator* is not specified, it defaults to [ascending](#ascending). For example:
 
 ```js
-const array = [{foo: 42}, {foo: 91}];
+const array = [{foo: 91}, {foo: 42}];
 d3.least(array, (a, b) => a.foo - b.foo); // {foo: 42}
 d3.least(array, (a, b) => b.foo - a.foo); // {foo: 91}
 d3.least(array, a => a.foo); // {foo: 42}
+d3.least(array); // {foo: 91}
```
mbostock, line 219: This isn’t a recommended pattern since the given array does not contain naturally orderable values, and an accessor/comparator is not specified. You will get the first non-nullish value this way, but we shouldn’t recommend this pattern. What gotcha/pattern were you trying to suggest here?

adebiyial: Hmm. I think you're correct. I was initially confused but not anymore.

And apologies for the late late response.

